PR #70, opened 2023-10-02
Finish extending transcript exon model

ielis:
Finish work on the model.

The main changes include:
- `Variant` has coordinates, functional annotations, and genotypes.
- use `Genotype` in the predicates. This is just a refactor, we must still finalize the predicate logic later
- update toy data and set genotypes for the variant @lnrekerle please notice the lines `49` and `54` in the doctest output of `tutorial.rst`. I think we had a bug here where we would count the same variant with different genotypes twice (e.g. `c.123C>T` with `homalt` and `het` genotypes would be counted as 2 distinct variants despite being just one variant.).
- update signature of `FunctionalAnnotator` to return a `typing.Sequence[TranscriptAnnotation]` instead of `Variant`
- setup `VariantTranscriptProteinArtist` stub


lnrekerle (approved):
Looks good! 
Let me know if you change the variant_string function in here, or I can in my next PR.

lnrekerle:
> * update toy data and set genotypes for the variant @lnrekerle please notice the lines `49` and `54` in the doctest output of `tutorial.rst`. I think we had a bug here where we would count the same variant with different genotypes twice (e.g. `c.123C>T` with `homalt` and `het` genotypes would be counted as 2 distinct variants despite being just one variant.).

Just noticed this comment. I can take a look at that after this is merged.

PR #74, opened 2023-10-03
Set `is_preferred` annotation

### LINKED ISSUE #73: Implement setting `is_preferred` field for `TranscriptAnnotation`
We have the `is_preferred` field on `TranscriptAnnotation` but we do not set it in `VepFunctionalAnnotator`.

## Definition of done
The value is parsed from the VEP response and the field is set.

ielis:
The PR adds a little refactor of the VEP functional annotator and implements parsing `is_preferred` field from the REST API response. The parsing logic is tested using two example responses (`_test_vep.py`) obtained by calling VEP endpoint as described in `test_data/vep_response/README.md`.

The VEP annotator (+tests) is moved into its own private package/file.

Closes #73


lnrekerle (approved):
Love this, looks good!

PR #77, opened 2023-10-10
HGVSVariantCoordinatesFinder

frehburg:
Dear @ielis,

We started implementing `HGVSVariantCoordinatesFinder` (inherits from `VariantCoordinateFinder`). However we ran into a few issues.

1. Should we just implement `VariantCoordinateFinder` or is it correct that we inherit from it and implement it specifically for HGVS?
2. Right now we are calling the Ensembl API from within Python and receiving the result as a dictionary instead of writing a json to file. I hope this works well with what you had in mind.
3. There were a few import issues, specifically regarding `Genotype`, `TranscriptInfoAware` and `TranscriptCoordinates`, so i commented the imports out. Further we couldn't return a `Genotype` object since we don't see any implementation of the class.

Please take a look, and let us know what you think.

P.S.: I think it would be simpler if we could be contributors on this repo directly instead of in our fork.


frehburg:
We could change it to work with this: https://rest.variantvalidator.org/VariantValidator/variantvalidator/hg38/NM_005912.3%3Ac.253A%3EG/mane_select?content-type=application%2Fjson, but i don't think it will make a big difference since we would be getting the same information I think.

ielis:
Hi @frehburg ,
I created a "stub" for you to implement at `genophenocorr.preprocessing._vep.VepHgvsVariantCoordinateFinder`, so there is no need to create `HGVSVariantCoordinatesFinder` in `_api`. The stub also has some comments and how I would go about implementing the required functionality.

So, wrt. the questions:
> Should we just implement VariantCoordinateFinder or is it correct that we inherit from it and implement it specifically for HGVS?

No, work on the stub.

> Right now we are calling the Ensembl API from within Python and receiving the result as a dictionary instead of writing a json to file. I hope this works well with what you had in mind.

Yeah, this is in line with my thoughts. No need to write a file, the code should, of course, parse JSON response in memory. The files that I created in `test_data` are designed to test the parsing logic only, and not the part that queries the server. I don't think we need to test the logic that gets us the response, but we need/should test the code that parses the response dict. Hence, the files and the unit tests.

> There were a few import issues, specifically regarding `Genotype`, `TranscriptInfoAware` and `TranscriptCoordinates`, so i commented the imports out. Further we couldn't return a `Genotype` object since we don't see any implementation of the class.

`Genotype` is an enum and the import should work in the context of the stub. Pls let me know if it does not work.

frehburg:
Hey @ielis please check the code on your machine. I have implemented everything to the best of my understanding and knowledge, however the unit tests still fail. The tests I inspected were due to assertions you hardcoded into the test cases. I think they might not be accurate to the actual responses.

E.g.:
**Assertion Error in `TestVepHgvsVariantCoordinateFinder.test_load_hgvs_MC4R_missense`:**
```
assert vc is not None
        assert vc.chrom == '18'
>       assert vc.start == 60_372_096
E       assert 60372097 == 60372096
E        +  where 60372097 = VariantCoordinates(chrom=18, start=60372097, end=60372097, ref=A, alt=G, change_length=1).start
```

**Actual data in `hgvs_MC4R_missense.json`:** (Same in the actual response when querying the server)
```
        "end": 60372097,
        "start": 60372097,
```

There are similar errors being thrown, however, I think everything should be accurate to the response.

ielis:
Hi @frehburg thanks a lot! I think this is more/less in line what should be done, so, it looks good in overall. I have a few comments.

## Major

### Coordinate systems

The unit tests are OK and you have to make them pass! 😎 A little more background may be needed. 
There are 2 types of coordinates in bioinformatics: 0-based and 1-based. Simply put, the 0-based coordinate system points to "spaces" between the bases and it is used e.g. in BED files. The 1-based coordinate system points to the bases, and it is used e.g. in VCF or GTF files.
 
To illustrate on a toy sequence:
```
0-based   0 1 2 3 4 5 6 7 8
1-based    1 2 3 4 5 6 7 8

seq        A C G T A C G T 
               👆
```

In 0-based coordinate system, the coordinates the highlighted `G` is located on $[2, 3]$, whereas in 1-based system, the coordinates would be $[3, 3]$.  

As another case, a *deletion* of the bases `GTA` would span the region with coordinates $[2, 5]$ (0-based) or $[3, 5]$ (1-based).

An *insertion* that adds `GGG` between `TA` would be located in the region with coordinates $[4, 4]$ (0-based) or $[5, 4]$ (1-based, !!!!).

Based on the above, you may understand why I *always* prefer the 0-based coordinate system and discourage usage of the 1-based. The nice properties of 0-based CS include:
- `start` is always before or at `end`
- length of the region is calculated simply as `end - start`
- the coordinates are used across computational community, e.g. to slice lists in python or creating substrings in Java

0-based coordinate system is *the* coordinate system. However, most people like to point fingers on stuff and I suspect that's why we have 1-based system. In *genophenocorr*, we can decide which one we use and I want to go with 0-based. If you have *no* social life at all, and want to know why I want to go with 0-based system, please read [docs on our Java library](https://github.com/exomiser/svart/blob/f2f1d8d9b8598fbcd7ba9ddcba9635ff81ea9a6c/src/main/java/org/monarchinitiative/svart/CoordinateSystem.java#L4) or [a note from dr. Edsger W. Dijkstra](https://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html).

### Use `variantvalidator.org` instead of VEP

Having said all this, there are few more extras. As a convenience, we would like to include the upstream base for INDELs. For instance, referring to the examples above, we would write the deletion as something like `seq:1-5CGTA>C` instead of `seq:2-5GTA>`. For the insertion, we want to get `seq:3-4T>TGGG` instead of `seq:4-4>GGG` (note the implied empty subparts of the `str`).

I looked at the VEP responses and I realized that the response does not include the upstream base. Unfortunately, this means we should use Variant validator [REST API](https://rest.variantvalidator.org) instead. Specifically, the `/VariantValidator/variantvalidator/{genome_build}/{variant_description}/{select_transcripts}` endpoint. The response includes the relevant data in the `primary_assembly_loci` field.

## Minor

- chrom is a `str` and not an `int`. Your code will explode on HGVS strings that represent variants on a gene that is located e.g. on `chrX`. There is a method on `GenomeBuild` that can get you a `Contig` for its name. Please inspect/debug the tests at `src/genophenocorr/model/genome/_test_builds.py`
- I beg you to disable/reduce strictness of the code formatter in your IDE. Your PR includes too many changes that are just reformatting of an existing code, which complicates the PR review.

## The game plan

Let me summarize the actions we need to move forward:

@ielis 
- I will fix the unit test JSON files to contain example responses from Variant Validator API
- I will move `VepHgvsVariantCoordinateFinder` to `VVHgvsVariantCoordinateFinder` in a separate file (`_vv` instead of `_vep`).
- I will push a fix for a bug that I found in the unit tests in one allele (`AAA` -> `AAAA`)

@frehburg 
- implement `VVHgvsVariantCoordinateFinder` and ensure the unit tests pass
- address the *minor* points above
- consider using a timeout in `requrests.get()`
- consider using `Strand.POSITIVE` instead of `Strand['POSITIVE']`

Thanks a lot! :)

frehburg:
Dear @ielis, thank you for your very thorough reply. I have been trained as a data scientist and am quite new to bio-/ medicine-informatics, so i really appreciate your explanation. I share your views on the indexing system (0).

I will work on this tomorrow, just already wanted to give a quick reply. Thanks again.

frehburg:
# Fixes TODO
## Major
- [ ] Fix coordinate systems
- [ ] Use variantvalidator.org instead of VEP
- [ ] implement VVHgvsVariantCoordinateFinder and ensure the unit tests pass

## Minor

- [ ] chrom is a str and not an int. Your code will explode on HGVS strings that represent variants on a gene that is located e.g. on chrX. There is a method on GenomeBuild that can get you a Contig for its name. Please inspect/debug the tests at src/genophenocorr/model/genome/_test_builds.py
- [ ] disable/reduce strictness of the code formatter in your IDE
- [ ] move VepHgvsVariantCoordinateFinder to VVHgvsVariantCoordinateFinder in a separate file (_vv instead of _vep).
- [ ] use a timeout in requrests.get()
- [ ] consider using Strand.POSITIVE instead of Strand['POSITIVE']
- [ ] consider compiling the regexp once and not per request

frehburg:
@ielis i set the timeout to 3 seconds for now, please adjust this as you see fit

frehburg:
Hey @ielis,

I am once again a bit stuck. As you can see on my branch, I am getting a response from vv at the moment, but this contains all of GRCh37, GRCh38, hg19, hg38, with different values for `'pos'`. However I couldn't find any start, end, or strand values. This might be caused by me passing `self._build.identifier` (evaluates to `'GRCh38.p13'` instead of any out of (GRCh37, GRCh38, hg19, hg38 ) as `genome_build` in `/VariantValidator/variantvalidator/{genome_build}/{variant_description}/{select_transcripts}`.

Maybe I should rather request from `/VariantValidator/tools/hgvs2reference/{hgvs_description}`

Could you please check this response?

```
{
  "NM_005912.3:c.253A>G": {
    "alt_genomic_loci": [],
    "annotations": {
      "chromosome": "18",
      "db_xref": {
        "CCDS": "CCDS11976.1",
        "ensemblgene": null,
        "hgnc": "HGNC:6932",
        "ncbigene": "4160",
        "select": "MANE"
      },
      "ensembl_select": false,
      "mane_plus_clinical": false,
      "mane_select": true,
      "map": "18q21.32",
      "note": "melanocortin 4 receptor",
      "refseq_select": true,
      "variant": "0"
    },
    "gene_ids": {
      "ccds_ids": ["CCDS11976"],
      "ensembl_gene_id": "ENSG00000166603",
      "entrez_gene_id": "4160",
      "hgnc_id": "HGNC:6932",
      "omim_id": ["155541"],
      "ucsc_id": "uc002lie.2"
    },
    "gene_symbol": "MC4R",
    "genome_context_intronic_sequence": "",
    "hgvs_lrg_transcript_variant": "LRG_1346t1:c.253A>G",
    "hgvs_lrg_variant": "LRG_1346:g.5672A>G",
    "hgvs_predicted_protein_consequence": {
      "lrg_slr": "LRG_1346p1:p.(S85G)",
      "lrg_tlr": "LRG_1346p1:p.(Ser85Gly)",
      "slr": "NP_005903.2:p.(S85G)",
      "tlr": "NP_005903.2:p.(Ser85Gly)"
    },
    "hgvs_refseqgene_variant": "NG_016441.1:g.5672A>G",
    "hgvs_transcript_variant": "NM_005912.3:c.253A>G",
    "primary_assembly_loci": {
      "grch37": {
        "hgvs_genomic_description": "NC_000018.9:g.58039330T>C",
        "vcf": {
          "alt": "C",
          "chr": "18",
          "pos": "58039330",
          "ref": "T"
        }
      },
      "grch38": {
        "hgvs_genomic_description": "NC_000018.10:g.60372097T>C",
        "vcf": {
          "alt": "C",
          "chr": "18",
          "pos": "60372097",
          "ref": "T"
        }
      },
      "hg19": {
        "hgvs_genomic_description": "NC_000018.9:g.58039330T>C",
        "vcf": {
          "alt": "C",
          "chr": "chr18",
          "pos": "58039330",
          "ref": "T"
        }
      },
      "hg38": {
        "hgvs_genomic_description": "NC_000018.10:g.60372097T>C",
        "vcf": {
          "alt": "C",
          "chr": "chr18",
          "pos": "60372097",
          "ref": "T"
        }
      }
    },
    "reference_sequence_records": {
      "lrg": "http://ftp.ebi.ac.uk/pub/databases/lrgex/pending/LRG_1346.xml",
      "protein": "https://www.ncbi.nlm.nih.gov/nuccore/NP_005903.2",
      "refseqgene": "https://www.ncbi.nlm.nih.gov/nuccore/NG_016441.1",
      "transcript": "https://www.ncbi.nlm.nih.gov/nuccore/NM_005912.3"
    },
    "refseqgene_context_intronic_sequence": "",
    "rna_variant_descriptions": null,
    "selected_assembly": "GRCh38.p13",
    "submitted_variant": "NM_005912.3:c.253A>G",
    "transcript_description": "Homo sapiens melanocortin 4 receptor (MC4R), mRNA",
    "validation_warnings": [
      "Invalid genome build has been specified. Automap has selected the default build (GRCh38)",
      "The current status of LRG_1346 is pending therefore changes may be made to the LRG reference sequence"
    ],
    "variant_exonic_positions": {
      "NC_000018.10": {
        "end_exon": "1",
        "start_exon": "1"
      },
      "NC_000018.9": {
        "end_exon": "1",
        "start_exon": "1"
      },
      "NG_016441.1": {
        "end_exon": "1",
        "start_exon": "1"
      }
    }
  },
  "flag": "gene_variant",
  "metadata": {
    "variantvalidator_hgvs_version": "2.2.0",
    "variantvalidator_version": "2.2.1.dev69+g7932ba3",
    "vvdb_version": "vvdb_2022_11",
    "vvseqrepo_db": "VV_SR_2022_11/master",
    "vvta_version": "vvta_2022_11_1"
  }
}
```

Needless to say, I am a bit confused

ielis:
Hi @frehburg
`pos` is the 1-based coordinate of the 1st base of the ref allele. You can derive `start` and `end` from the `pos`.

E.g. for `grch38`:
```json
"hg38": {
    "hgvs_genomic_description": "NC_000018.10:g.60372097T>C",
    "vcf": {
        "alt": "C",
        "chr": "chr18",
        "pos": "60372097",
        "ref": "T"
    }
}
```

- `start=60_372_096`
- `end=60_372_097`

Deriving `start` and `end` is relatively straightforward for SNPs, but it can be a bit more elaborate for INDELs.. Try to get the unit tests that test the SNPs first and let me know if you run into troubles with INDELs.

P.S.: VCF coordinates always use `Strand.POSITIVE`.
P.S2: Please use the `genophenocorr.model.genome.GenomeBuild.identifier` property to figure out which sub-dict should be extracted from the VV's response. It's not an exact match, but it's fairly close and there are just two values in `genophenocorr/model/genome/_builds.py`

ielis:
Genophenocorr takes a phenopacket and phenopacket can have variant data encoded either as VCF record, or an expression, such as HGVS:

For instance:
```json
"variationDescriptor": {
  "id": "variant id",
  "expressions": [{
    "syntax": "hgvs",
    "value": "NM_007129.3:c.1069C>G"
  }],
  "vcfRecord": {
    "genomeAssembly": "GRCh38",
    "chrom": "NC_000013.11",
    "pos": "99983133",
    "ref": "C",
    "alt": "G"
  },
  "moleculeContext": "unspecified_molecule_context",
  "allelicState": {
    "id": "GENO:0000135",
    "label": "heterozygous"
  }
}
```

In cases like this, genophenocorr uses VCF coordinates to ingest the variant. 

However, Adam's & FIlip's phenopackets lack VCF coordinates. All they have is an HGVS string. To fix this, I want to add a new component into the genophenocorr's framework, to translate HGVS -> `VariantCoordinates`. That's what Filip has been working on.

After we're done, we'll plug the new component into the phenopacket parsing logic, to use VCF first and fall back to an HGVS expression.

pnrobinson:
The code I had goes from HGVS to genomic - it works! 
 :-0

ielis:
Yeah, it should work here too.. We'll get there.

frehburg:
Dear Daniel,

I have implemented it to the point such that the tests for `hgvs_MC4R_missense.json`, `hgvs_MC4R_dup.json` and `hgvs_SURF2_ins.json` are passing. All properties but start and end are calculated correctly for all tests.

I don't understand the test case for `hgvs_SURF2_dup.json`.

Why are the correct values `start=133357812, end=133357813` if `pos=133357810`. (this is a dup hgvs string with `ref=C, alt=CCAT`)

`hgvs_MC4R_dup.json` also is a dup hgvs string with `ref=T, alt=TAGC` and here the math works out such that `pos == end`: `pos=60372346, start=60372345, end=60372346`.

In both cases the `ref` and `alt` alleles are of length `1` and `4` respectively with the first base coinciding. I don't understand what produces the difference in behavior. 

Maybe calculating the start and end position of the mutation requires some additional information?

I suppose also the type of mutation plays a role.

Further the test for `hgvs_SURF2_del.json` fails.

These are problems I am very certain I could usually solve in a few minutes, it it mainly the domain knowledge that I am missing here. I would be happy to learn more :)

Here the code which I have right now to calculate the start, end and change length:

```
def _extract_0based_mutation_range_from_1based_pos(self, pos: int, ref: str, alt: str) -> typing.Tuple[int, int, int]:
        """
        :return: 0-based start, end and change_length
        """
        print(f'{pos=}')
        change_length = len(alt) - len(ref)
        end = pos
        start = pos - len(ref)

        return start, end, change_length
```

Looking forward to your reply! - F

frehburg:
Hey @ielis,

all tests now pass. 

I set the timeout to 10s, do you find this appropriate? we can change it otherwise. You can find it in line 26 of `_vv.py`.

Please check if everything else (formatting, documentation, etc) is up to your expectations and then you could merge this.

PR #78, opened 2023-10-10
Moved VariantEffects so it's using the Enum

lnrekerle:


ielis (changes requested):
Hi @lnrekerle I think this is great!

I pushed a few little tweaks and there is one more thing that I'd like to ask you to do. Please see the comment below, regarding handling missing `VariantEffect` enum members.

ielis (commented):
So, as far as I can see, `5` and `3` are the only numbers we can encounter in variant effect names. This is because the "direction" of DNA depends on moieties bound to 3' and 5' carbons in ribose. So, correct me if I'm wrong, but I think these are the only numbers we can reasonably expect to get in the context of `VariantEffect` and, therefore, having `digit_word_map` and replacing `0` with `ZERO` seems reeeally odd..

I think we should handle this with two explicit *if/else* tests:
```python
if effect == '5_PRIME_UTR_VARIANT':
  pass
elif effect == '3_PRIME_UTR_VARIANT':
  pass
else:
  # use VariantEffect[effect.upper()] and log warning if things go wrong.
  pass
```

BTW, `self._logging` is not callable.

src/genophenocorr/preprocessing/_vep.py
```diff
@@ -80,6 +80,18 @@ def annotate(self, variant_coordinates: VariantCoordinates) -> typing.Sequence[T
 
         return annotations
 
+    def _parse_variant_effect(self, effect: str) -> typing.Optional[VariantEffect]:
+        if effect.upper() == "5_PRIME_UTR_VARIANT":
```
ielis, line 84: I have two minor suggestions:
-  please refactor such that you do the conversion to uppercase at most once per method body instead of 3 times in the worst case.
- logging, as currently written, will construct the string unconditionally, regardless of the set log level threshold (e.g. even if the user sets the threshold to see ERRORs only). This is because you're submitting the f-string which should be avoided when logging a statement. Instead, we should do something like this:
  ```python
  self._logging.warning('Variant effect %s was not found ...', effect)
  ```
  Using this form, the logging framework will only create the final string if the logging threshold is low enough (WARN or lower in this case).

ielis (changes requested):
Hi @lnrekerle 
I'd like to ask you to do 2 more updates, see the comment below. After that, I think this is ready to go.

ielis (approved):
Looks good, thanks!

Issue #81, opened 2023-10-19
as_string() does not exist


pnrobinson:
which means that this lin in ``_vep.py`` is throwing an additional error
```
self._logging.error(f"Expected a result but got an Error for variant: {variant_coordinates.as_string()}")
```
VariantCoordinates has a __str__ method, which means we can write simply
```
self._logging.error(f"Expected a result but got an Error for variant: {variant_coordinates}")
```

pnrobinson:
Fixed by https://github.com/monarch-initiative/genophenocorr/pull/82

pnrobinson:
fixed

PR #82, opened 2023-10-19
https://github.com/monarch-initiative/genophenocorr/issues/81

pnrobinson:


ielis (approved):
Hi @pnrobinson sorry we removed `as_string`, and, apparently, forgot to fix one use. The `as_string` was replaced by `variant_key` - a method with the same purpose and a better name.

I added a few fixes that caught my attention. Pls merge when convenient.

PR #85, opened 2023-10-19
Created Cohort Viewer that shows Excluded Patients

lnrekerle:


src/genophenocorr/model/_variant_effects.py
```diff
@@ -74,3 +74,11 @@ def curie(self) -> str:
 
     def __str__(self) -> str:
         return self.name.lower()
+
```
ielis, line 77: In Python, `Enum` members are singletons, therefore, `__eq__` and `__hash__` should not be overriden.

src/genophenocorr/model/_protein.py
```diff
@@ -242,6 +246,18 @@ def motifs(self) -> typing.Iterable[ProteinFeature]:
...
+        affected_features = set()
+        for feat in self.protein_features:
+            if feat.info.start is None or feat.info.end is None:
+                print(f"{feat.info.name} has no start and end info")
+                continue
+            if feat.info.start <= var_start <= feat.info.end:
+                affected_features.add(feat.to_string())
+            elif feat.info.start <= var_end <= feat.info.end:
+                affected_features.add(feat.to_string())
+        return affected_features
```
ielis, line 259: `set` is not a `Sequence`. Moreover, you are not returning a `Set[ProteinFeature]` but a `Set[str]`.

src/genophenocorr/model/_protein.py
```diff
@@ -242,6 +246,18 @@ def motifs(self) -> typing.Iterable[ProteinFeature]:
...
         return filter(lambda f: f.feature_type == FeatureType.MOTIF, self.protein_features)
 
+    def get_features_variant_overlaps(self, var_start: int, var_end: int) -> typing.Sequence[ProteinFeature]:
+        affected_features = set()
+        for feat in self.protein_features:
+            if feat.info.start is None or feat.info.end is None:
+                print(f"{feat.info.name} has no start and end info")
+                continue
+            if feat.info.start <= var_start <= feat.info.end:
+                affected_features.add(feat.to_string())
```
ielis, line 256: `ProteinFeature` has no `to_string` method but `SimpleProteinFeature` has it. This code will fail when `FeatureInfo` is provided with a different `ProteinFeature` implementation, which is certainly a legal thing to do.

This bug as well as the one below are glaring into my eyes in PyCharm thanks to the typing annotation and the PyCharm linter. Please consider setting up a Python linter in Code, or switching to PyCharm. :roll_eyes:

src/genophenocorr/model/_protein.py
```diff
@@ -141,6 +142,9 @@ def feature_type(self) -> FeatureType:
         """
         return self._type
 
+    def to_string(self) -> str:
```
ielis, line 145: Unless I don't see something, you're using this method in a context of getting a unique item collection. However, the method does not use the entire state of the instance, so the logic for getting unique elements based on this method is broken. 

Assessing uniqueness is responsibility of `__hash__`. So, I recommend removing this method and relying on `__hash__` instead, if unique elements is what you're after.

src/genophenocorr/preprocessing/_phenopacket.py
```diff
@@ -172,7 +172,7 @@ def _add_phenotypes(self, pp: Phenopacket) -> typing.Sequence[Phenotype]:
         for hpo_id in pp.phenotypic_features:
             hpo_id_list.append((hpo_id.type.id, not hpo_id.excluded))
         if len(hpo_id_list) == 0:
-            self._logger.warning(f'Expected at least one HPO term per patient, but received none for patient {pp.id}')
+            #self._logger.warning(f'Expected at least one HPO term per patient, but received none for patient {pp.id}')
```
ielis, line 175: Commented code is a sign of an immature code.

src/genophenocorr/preprocessing/_uniprot.py
```diff
@@ -20,6 +20,11 @@ def __init__(self):
         """Constructs all necessary attributes for a UniprotProteinMetadataService object
         """
         self._logger = logging.getLogger(__name__)
+        self._logger.setLevel(logging.INFO)
```
ielis, line 23: This logging configuration must be removed because setting up logging is the user's responsibility, not ours.

Note, you can use this to quickly setup logging in notebook or a script:

```python
import hpotk
hpotk.util.setup_logging()
```

src/genophenocorr/preprocessing/_uniprot.py
```diff
@@ -64,8 +68,8 @@ def annotate(self, protein_id: str) -> typing.Sequence[ProteinMetadata]:
                     self._logger.warning(f"No features for {protein_id}")
                 protein_list.append(ProteinMetadata(protein_id, protein_name, all_features_list))
             else:
-                self._logger.warning(f"ID {protein_id} did not match")
-        self._logger.warning(f'Protein ID {protein_id} got {len(protein_list)} results')
-
+                self._logger.warning(f"UniProt did not return a protein ID that matches the ID we searched for: {protein_id} not in {unis}")
+        if len(protein_list) > 1:
```
ielis, line 72: People are usually interested in `logging.INFO` messages. However, here we'll produce an INFO message for each successful Uniprot query, which will, hopefully, be >99% of time. It would be better to make this a debug statement, or just not log anything.

src/genophenocorr/preprocessing/_uniprot.py
```diff
@@ -64,8 +68,8 @@ def annotate(self, protein_id: str) -> typing.Sequence[ProteinMetadata]:
                     self._logger.warning(f"No features for {protein_id}")
                 protein_list.append(ProteinMetadata(protein_id, protein_name, all_features_list))
             else:
-                self._logger.warning(f"ID {protein_id} did not match")
-        self._logger.warning(f'Protein ID {protein_id} got {len(protein_list)} results')
-
+                self._logger.warning(f"UniProt did not return a protein ID that matches the ID we searched for: {protein_id} not in {unis}")
```
ielis, line 71: We recently went through the best practices regarding parameterizing logging statements...

ielis (changes requested):
Hi @lnrekerle see the comments below.

src/genophenocorr/model/_protein.py
ielis: It looks like MyPy linter can catch these bugs:

image

image

src/genophenocorr/preprocessing/_uniprot.py
lnrekerle: ` if len(protein_list) > 1:
     self._logger.info(f'UniProt found {len(protein_list)} results for ID {protein_id}')`

Think this is the line you are referring to, and this will actually rarely run since I have almost never gotten more than one UniProt result for a given ID.

ielis: OK, let's leave it here if it occurs rarely.

ielis (changes requested):
Hi @lnrekerle 
looks good, thanks!

I fixed one bug in the code that searched for overlap between protein coordinates and a protein feature. I added few new methods to `Region` to do that.

Otherwise, there is one thing that I'd like to point out (see also the comment below). You should use type annotations to write your code. When you look into `Region`, the `start` and `end` properties return `int`. Not an `int` or `None`. Therefore, testing if the coordinate is `None` is redundant and wrong.

This is the point of using the type annotations. Please work on adding them into your repertoire.

src/genophenocorr/model/_protein.py
```diff
@@ -242,6 +246,18 @@ def motifs(self) -> typing.Iterable[ProteinFeature]:
         """
         return filter(lambda f: f.feature_type == FeatureType.MOTIF, self.protein_features)
 
+    def get_features_variant_overlaps(self, var_start: int, var_end: int) -> typing.Set[ProteinFeature]:
+        affected_features = set()
+        for feat in self.protein_features:
+            if feat.info.start is None or feat.info.end is None:
```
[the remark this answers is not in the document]
lnrekerle: Made the change!

ielis (approved):
Hi @lnrekerle looks good, thank you!

While looking at the `ProteinMetadata.get_features_variant_overlaps()`, I went ahead and changed the signature to take a `Region` instead of a pair of `int`s. Thanks to this we have few things less to worry about in the downstream code. 

First, we do not have to worry about an invalid state in the method where `start` comes after `end`. We would have to check this in the function to be sure but now we don't have to, because coordinate "sanity" is one of `Region`'s invariants.

Next, we don't have to worry about coordinate system in the downstream code, e.g. in predicates. Thanks to using `Region`, we can be sure that we're comparing apples with apples, when e.g. checking the values in `ProtFeatureTypePredicate` or `ProtFeaturePredicate` predicates. In other words, we are certain that the coordinates are in the same coordinate system and we are not prone to off-by-one bugs, a very frequent bugs in bioinformatics due to mishandling of the coordinate systems.

With these, I think we're good to merge, thanks a lot!

PR #88, opened 2023-10-23
Streamline predicate/analysis workflow

### LINKED ISSUE #87: Streamline predicate/analysis workflow
Currently, all `PolyPredicate` implementations return these values as available categories:
```
HETEROZYGOUS, HOMOZYGOUS, NO_VARIANT
```

However, the categories do not make intuitive sense in context of several predicates. 

For instance `VariantEffectPredicate` should test if the patient has a variant with the `query: VariantEffect` on the given transcript or not. So, the predicate should in fact be a simple predicate (not poly) and return `True`/`False` for a patient that has or has not such variant.

`ExonPredicate` is another example. The predicate should return `True`/`False` based on if the variant affects the exon in question or not.

Having more accurate predicate categories, the `CohortAnalyzer` should be able to treat the predicates the same, which simplifies the code. It should be easy to  apply an additional logic. For instance, filter out the patients that do not meet the expected mode of inheritance, count the groups, do statistical testing, apply multiple-testing correction, etc..

ielis:
Fixes #87 , #92 

Depends on #94


ielis:
@lnrekerle 

I'm proposing a revamp to the `CohortAnalyzer`.

The `CohortAnalyzer` is an abstraction - a promise what `CohortAnalyzer` can do for the user. To get `CohortAnalyzer` we use a similar pattern to configuring `PhenopacketPatientCreator`. There is a config method that will give you `CohortAnalyzer`:

```python
from genophenocorr.analysis import configure_cohort_analysis

analysis = configure_cohort_analysis(cohort, hpo)
```

You'll get an analysis with default options. If you want to tweak the options, build the `CohortAnalysisConfiguration`:

```python
from genophenocorr.analysis import CohortAnalysisConfiguration

configuration = CohortAnalysisConfiguration.builder()
  .include_sv(True)
  .pval_correction('fdr_bh')
  .build()

analysis = configure_cohort_analysis(cohort, hpo, configuration)
```

Then we run the analysis, e.g. to compare MISSENSE vs others:

```python
from genophenocorr.model import VariantEffect
from genophenocorr.analysis.predicate import BooleanPredicate

results = analysis.compare_by_variant_effect(VariantEffect.MISSENSE_VARIANT, tx_id='NM_1234.5')
result_df = results.summarize(hpo, BooleanPredicate.YES)
result_df.head()
```

We get `results`, a container with a lot of data. We call summarize to prepare a data frame with phenotypes vs. genotypes, ordered by corrected p values.

Note that we provide `BooleanPredicate.YES` to show genotype-phenotype correlation for *present* HPO terms, not for not-present (we would use `BooleanPredicate.NO` to show those).

This is what the PR adds. Thanks to the changes, we have a general framework for applying genotype and phenotype predicates and showing the results.

Please check out the code, try it out and we can discuss in greater detail the next time.

ielis:
Now, with the `develop` merged into the PR branch, we should be OK to move forward with this PR if the code looks good.

Issue #83, opened 2023-10-19
better error messages


pnrobinson:
When we create cohorts, variants in phenopackets are annotated and run through the _vep.py code.
One difficulty is that if there is an error, the ingest is killed, but the error message is not enough to know which phenopacket the error is coming from, e.g. 

```
Expected a result but got an Error for variant: VariantCoordinates(chrom=9, start=127665287, end=127665288, ref=A, alt=G, change_length=0)
```
(note -- you will only see this error message after merge the PR I just made).
This is very difficult because (in this case) the original variant coordinates are hg19 and the above coordinates are hg38, and it took me about 15 minutes to track this down.

It would be better to catch this Exception and then rethrow it with the patient or phenopacket ID.  I am not sure of the best place to add this?

ielis:
Hi @pnrobinson I think genome build mismatch is a situation that the code should handle. We have enough context to check if the builds match, see #84 for verbose info. So, I or Lauren will eventually fix this. 

The only outstanding question is if we want to fail the entire ingest due to a bug like this. Currently, the framework *must* provide `VariantCoordinates` for a `GenomicInterpretation`, so, the only possible way to fail is to raise and explode at the parsing site. The upstream code that handles the phenopacket can either catch this exception or explode too. 

Should we halt the entire ingest and point the user to the offending phenopacket or log the warning about invalid variant, swallow the exception, and ignore the variant? I think we should force the user to provide good input for now and defer the ergonomics to a later stage. What do you think?

pnrobinson:
Currently, the error message does not tell you the name of the erroneous phenopacket. It would make it much easier to include the name of the phenopacket in the error message for every time that we terminate the program like this -- otherwise it will be nearly impossible for others to use this package!

Issue #52, opened 2023-09-20
user-friendly error messages


pnrobinson:
When loading phenopackets like this:
```
patient = patient_creator.create_patient(pp)
```
I get these error warnings:
```
No proteins found for ID NP_001361354.1. Please verify refseq ID.
No proteins found for ID NP_001361355.1. Please verify refseq ID.
No proteins found for ID NP_001361356.1. Please verify refseq ID.
ID NP_002608.1 did not match
ID NP_002608.1 did not match
ID NP_002608.1 did not match
Protein ID NP_002608.1 got 1 results
ID NP_008964.3 did not match
ID NP_008964.3 did not match
Protein ID NP_008964.3 got 1 results
ID NP_722540.1 did not match
ID NP_722540.1 did not match
ID NP_722540.1 did not match
Protein ID NP_722540.1 got 1 results
```
This looks like something really bad because the background is read. However, the warnings are redundant and maybe we do not care about some of them. It would be better to create one message about whether we have successfully created the objects and try to decide if we think it is an error that UniProt cannot find NP_001361354.1 (if that is what is going on). I do not think this should be an error

Issue #46, opened 2023-09-14
Update the notebooks


ielis:
Update the notebooks with the most recent code.

lnrekerle:
- [ ] KBG
- [ ] SUOX
- [ ] MAPK8IP3
- [ ] RPGRIP1

ielis:
### TODO

- make a folder with 3 notebooks
- the folder has a subfolder `phenopackets`
  - `KBG`
  - `SUOX`
  - ...

TODO - finish

PR #96, opened 2023-10-27
Error messages

### LINKED ISSUE #108: Erroneous error while loading individuals to cohort. Proteins are not expected to have coordinates for splice variants
This is expected.
I think we need to slightly refactor the way the analysis starts. We need to give the analysis a reference transcript (not a reference protein). 
It also does not make sense to test multiple proteins (or transcripts) at once. For any conceivable clinical analysis, people focus on one transcript, which in almost all cases is (or should be) the MANE or ClinVar transcript. Having this much choice will engender confusion and not bring benefits!

```
Missing start/end coordinate for NM_001330437.2:c.1093-1G>T on protein NP_001317366.1
Missing start/end coordinate for NM_001374625.1:c.1090-1G>T on protein NP_001361554.1
Missing start/end coordinate for NM_002834.5:c.1093-1G>T on protein NP_002825.3
Missing start/end coordinate for NM_080601.3:c.1093-1G>T on protein NP_542168.1
Missing start/end coordinate for NM_001330437.2:c.643-2A>C on protein NP_001317366.1
```

lnrekerle:
#83 #52 
This PR will update error messages and create logging files for Warnings/Info that will be easier for users to understand. 

edit: 
Also will close #108


lnrekerle:
I think this is good for now, we will obviously want to add more errors as we continue.

src/genophenocorr/analysis/_commie.py
```diff
@@ -106,15 +106,22 @@ def __init__(self, cohort: Cohort,
...
                  include_sv: bool = False,
                  p_val_correction: typing.Optional[str] = None,
-                 min_perc_patients_w_hpo: typing.Union[float, int] = .1):
+                 min_perc_patients_w_hpo: typing.Union[float, int] = .1,
+                 recessive: bool = False):
         if not isinstance(cohort, Cohort):
             raise ValueError(f"cohort must be type Cohort but was type {type(cohort)}")
 
         self._logger = logging.getLogger(__name__)
+        handler = logging.FileHandler(f"{__name__}.log", mode='w')
```
ielis, line 115: Logging must not be configured here. We must not make decisions on behalf of the user. The lines 115-118 must be removed.

src/genophenocorr/analysis/_commie.py
```diff
@@ -106,15 +106,22 @@ def __init__(self, cohort: Cohort,
                  missing_implies_excluded: bool = False,
                  include_sv: bool = False,
                  p_val_correction: typing.Optional[str] = None,
-                 min_perc_patients_w_hpo: typing.Union[float, int] = .1):
+                 min_perc_patients_w_hpo: typing.Union[float, int] = .1,
+                 recessive: bool = False):
```
ielis, line 110: Whether we do a recessive, dominant, or other analysis is not a concern of `CohortAnalysis`, and, therefore, does not belong here. Checking the mode of inheritance and binning the patients accordingly is a responsibility of the predicates.  

We should discuss this during Zoom, but I think all changes related to `recessive` should be removed.

src/genophenocorr/analysis/_config.py
```diff
@@ -31,11 +44,13 @@ class CohortAnalysisConfiguration:
     def __init__(self, missing_implies_excluded: bool,
                  pval_correction: typing.Optional[str],
                  min_perc_patients_w_hpo: float,
-                 include_sv: bool):
+                 include_sv: bool, 
+                 recessive: bool):
```
ielis, line 48: `recessive` does not belong here too. It depends on the predicate.

src/genophenocorr/analysis/_config.py
```diff
@@ -87,10 +111,15 @@ class CohortAnalysisConfigurationBuilder:
 
     def __init__(self):
         self._logger = logging.getLogger(__name__)
+        handler = logging.FileHandler(f"{__name__}.log", mode='w')
```
ielis, line 114: Remove logging config.

src/genophenocorr/analysis/_config.py
```diff
@@ -108,16 +137,20 @@ def pval_correction(self, pval_correction: typing.Optional[str]):
         """
         Set `pval_correction` option.
         """
-        # TODO - check admissible values
-        self._pval_correction = pval_correction
+        if pval_correction in P_VAL_OPTIONS:
```
ielis, line 140: This is wrong. The user should be allow to turn off the p value correction by setting `None`. However, since `None` is not a member of `P_VAL_OPTIONS`, the correction will not be turned off. Instead of that, the user will see a deceiving logging message 
```
Ignoring invalid `pval_correction` value None. Not doing p-value correction.
```
when in fact `bonferroni` will be done.

You must fix this to ensure that only invalid `str`s are complained about and we can turn the correction off.

src/genophenocorr/analysis/_config.py
```diff
@@ -108,16 +137,20 @@ def pval_correction(self, pval_correction: typing.Optional[str]):
...
+            self._logger.warning('Ignoring invalid `pval_correction` value %s. Not doing p-value correction.', pval_correction)
         return self
 
     def min_perc_patients_w_hpo(self, min_perc_patients_w_hpo: float):
         """
         Set `min_perc_patients_w_hpo` option.
         """
-        # TODO - check float in range [0,1)
-        self._min_perc_patients_w_hpo = min_perc_patients_w_hpo
+        if min_perc_patients_w_hpo > 1 or min_perc_patients_w_hpo <= 0:
```
ielis, line 150: I think we should first check if the argument is `float`. Then we can check the rest.

src/genophenocorr/analysis/_config.py
```diff
@@ -108,16 +137,20 @@ def pval_correction(self, pval_correction: typing.Optional[str]):
...
+            self._logger.warning("min_perc_patients_w_hpo must be greater than 0 and at most 1. Using default of 0.1")
```
ielis, line 151: Showing logging messages such as this can be helpful but also deceiving if not done right. 

If we update the default value, in this case on line 120, we will almost *surely* forget to update the logging message. I think you should parameterize the logging message to print the current value of `self._min_perc_patients_w_hpo`. Then it will be futureproof.

src/genophenocorr/preprocessing/_phenopacket.py
```diff
@@ -29,6 +29,10 @@ class PhenopacketVariantCoordinateFinder(VariantCoordinateFinder[GenomicInterpre
     def __init__(self, build: GenomeBuild,
                  hgvs_coordinate_finder: VariantCoordinateFinder[str]):
         self._logger = logging.getLogger(__name__)
+        handler = logging.FileHandler(f"{__name__}.log", mode='w')
```
ielis, line 32: Remove

src/genophenocorr/preprocessing/_phenopacket.py
```diff
@@ -140,6 +144,10 @@ def __init__(self, build: GenomeBuild,
                  var_func_ann: FunctionalAnnotator,
                  hgvs_coordinate_finder: VariantCoordinateFinder[str]):
         self._logger = logging.getLogger(__name__)
+        handler = logging.FileHandler(f"{__name__}.log", mode='w')
```
ielis, line 147: Remove

src/genophenocorr/preprocessing/_phenopacket.py
```diff
@@ -181,15 +189,18 @@ def _add_variants(self, sample_id: str, pp: Phenopacket) -> typing.Sequence[Vari
             for genomic_interp in interp.diagnosis.genomic_interpretations:
                 vc, gt = self._coord_finder.find_coordinates(genomic_interp)
                 if "N" in vc.alt:
-                    self._logger.warning(f'Patient {pp.id} has unknown alternative variant {vc.alt} and will not be included.')
+                    self._logger.warning('Patient %s has unknown alternative variant %s, this variant will not be included.', pp.id, vc.variant_key)
                     continue
                 tx_annotations = self._func_ann.annotate(vc)
+                if tx_annotations is None:
+                    self._logger.error("Patient %s has an error with variant %s, this variant will not be included.", pp.id, vc.variant_key)
```
ielis, line 196: I think this should be a `warning`, not an `error`.

src/genophenocorr/preprocessing/_vep.py
```diff
@@ -77,10 +81,12 @@ def annotate(self, variant_coordinates: VariantCoordinates) -> typing.Sequence[T
         """
         response = self._query_vep(variant_coordinates)
         annotations = []
+        if response is None:
+            self._logger.error('VEP did not finish successfully.')
```
ielis, line 85: `_query_vep` always logs an error before returning `None`. This is, therefore, redundant and should be removed. If you feel the user need more context, then you should update the logging in the `_query_vep` method instead.

ielis (changes requested):

src/genophenocorr/preprocessing/_phenotype.py
```diff
@@ -48,8 +52,9 @@ def create_phenotype(self, term_ids: typing.Iterable[typing.Tuple[str, bool]]) -
         for term_id, observed in term_ids:
             term = self._hpo.get_term(term_id)
             if term is None:
-                raise ValueError(f'Term ID {term_id} is not present in HPO v{self._hpo.version}')
-            terms.append((term, observed))
+                self._logger.warning("Term %s cannot be found in HPO version %s. It will be ignored.", term_id, self._hpo.version)
```
ielis, line 55: I think this warning is important but not too useful because it lacks context. I bet the user will be interested in fixing this. However, to do that, he must know about the sample ID. 

However, we do not have the sample ID around here. We may need to restructure this code. I'll make a ticket.

src/genophenocorr/analysis/_config.py
ielis: EDIT, actually now I see that `None` in fact IS in `P_VAL_OPTIONS`, so I was not entirely right above.

However, the reporting of what will be done if the user provides an invalid `pval_correction` must still be fixed. The warning message claims that the correction will not be done, but that actually depends on the previous `self._pval_correction` value, which is `bonferroni` by default.

lnrekerle: Fixed logging message.

src/genophenocorr/analysis/_config.py
lnrekerle: Added a float check

src/genophenocorr/preprocessing/_phenotype.py
lnrekerle: Leaving this for now, but agreed a ticket to make this more useful would be helpful.

src/genophenocorr/analysis/_commie.py
lnrekerle: Removed

src/genophenocorr/analysis/_config.py
lnrekerle: Fixed this!

src/genophenocorr/analysis/_config.py
```diff
@@ -140,15 +124,21 @@ def pval_correction(self, pval_correction: typing.Optional[str]):
         if pval_correction in P_VAL_OPTIONS:
             self._pval_correction = pval_correction
         else:
-            self._logger.warning('Ignoring invalid `pval_correction` value %s. Not doing p-value correction.', pval_correction)
+            self._logger.warning('Ignoring invalid `pval_correction` value %s. Using default "bonferroni" correction.', pval_correction)
```
ielis, line 127: @lnrekerle I hate myself for this but I cannot stay quiet. This is still not right. Consider this situation:

```python
config = CohortAnalysisConfiguration.builder()
  .pval_correction('sidak')
  .pval_correction('EVIL HACK')
  .build()
```
The code above will log `EVIL_HACK` and advertise falling back to `bonferroni`, while `sidak` is actually what will be done!
So, you must log the actual value instead of `bonferroni`.

---
EDIT: I think this will also apply to other changes in this builder. Please revise accordingly.

src/genophenocorr/preprocessing/_phenotype.py
ielis: :+1: #113

ielis (commented):
Hi @lnrekerle thanks a lot for the updates.

On the lower level, I think the changes look good except for that one thing in the builder that I mention below. Please consider changing that, because I think this is a bug that can deceive the users and cause confusion.

On the higher level, I would like to see how the logging looks like in an example analysis. Therefore, let's please wait with merging until you're finished with #46. Then we can see if the logging looks reasonable.

Thank you!

src/genophenocorr/analysis/_config.py
lnrekerle: I think the new messages are better. Let me know what you think.

ielis: Looks good, thanks!

src/genophenocorr/preprocessing/_vep.py
```diff
@@ -134,8 +135,11 @@ def _process_item(self, item: typing.Dict) -> typing.Optional[TranscriptAnnotati
             # if we see a lot of these warnings popping out.
```
ielis, line 135: Please remove the comments, now we know that this happens.

src/genophenocorr/preprocessing/_vep.py
```diff
@@ -8,6 +8,7 @@
 from genophenocorr.model.genome import Region
 from ._api import FunctionalAnnotator, ProteinMetadataService
 
+SPLICE_EFFECTS = [VariantEffect.SPLICE_ACCEPTOR_VARIANT, VariantEffect.SPLICE_DONOR_VARIANT, VariantEffect.SPLICE_DONOR_5TH_BASE_VARIANT, VariantEffect.SPLICE_POLYPYRIMIDINE_TRACT_VARIANT]
```
ielis, line 11: As a rule of thumb, you should always try to keep a variable declaration as close to the usage as possible.

In this case, the best place to put the `SPLICE_EFFECTS` is likely a class field, because it is a static variable.

Please consider moving `SPLICE_EFFECTS` closer to usage.

src/genophenocorr/preprocessing/_vep.py
```diff
@@ -134,8 +135,11 @@ def _process_item(self, item: typing.Dict) -> typing.Optional[TranscriptAnnotati
             # if we see a lot of these warnings popping out.
             # Note that Lauren's version of the code had a special branch for missing start, where she set the variable
             # to `1` (1-based coordinate).
-            self._logger.warning('Missing start/end coordinate for %s on protein %s', hgvsc_id, protein_id)
-            protein_effect = None
+            if any(ve in var_effects for ve in SPLICE_EFFECTS):
+                protein_effect = None
```
ielis, line 139: You're repeating yourself here. Ctrl+C Ctrl+V in a small scale. It would be the best to know how to extract the common case so that you do not repeat yourself and the code is, in result, shorter.

ielis (approved):
Hi @lnrekerle ,

I have some "cosmetic" suggestions. Please address them, if you wish so, and then feel free to merge.

Thanks!

PR #107, opened 2023-11-30
Used tqdm to create a progress bar

### LINKED ISSUE #72: load_phenopacket_folder
1. add tqdm to for loop to show progress
2. make sure we only query VEP once per distinct variant

lnrekerle:
closes #72 

Creates a progress bar that will show how many patients have been completed out of how many total when running load_phenopacket_folder. Looks better when the PR Error messages is merged as well, print statements throw off the look of the bar, but can be pushed before that PR.


ielis (approved):
Looks great, thank you!

PR #122, opened 2023-12-19
Run notebooks in CI

### LINKED ISSUE #86: Add action that runs notebooks
Add GitHub action that runs notebooks to ensure the notebooks and the code won't get out of sync.

The notebooks should choose stable web-based HPO to ensure we don't require to keep HPO in the repo. However, do not choose latest HPO to avoid non-determinism.

The action should be run on PRs against `main` and `develop`.

ielis:
Adds GitHub action for running selected notebooks. Not all notebooks are run to keep the execution time in reasonable boundaries.

On top of the main item, the PR improves logging by not complaining about missing protein annotations for additional classes of non-coding variants (see the `NONCODING_EFFECTS` attribute of `VepFunctionalAnnotator`.

Furthermore, the PR ensures we sort the phenotype features by first the corrected p value, and then the uncorrected value. Previously, the data frame was sorted by the corrected values only, which is sub-optimal if all values are equal to `1.0`.

Last, I went through the notebooks and simplified the code, while preserving as much of the original content as possible. In result, all notebooks should be runnable. However, the results still need to be worked on to ensure we ask the proper questions - the questions from the respective papers.

Fixes #86


lnrekerle (approved):
Looks good!

PR #123, opened 2023-12-19
Minor updates

ielis:
A PR with few minor ad hoc fixes:

- ignore VS Code workspace files
- update dependency version bounds. We probably want to be a bit more permissive on `tqdm` and set up upper bounds for other packages
- set `tool.setuptools` project attribute to point to `src` folder


lnrekerle (approved):
Looks good!

PR #128, opened 2024-01-11
Handle phenotype creator errors

ielis:
Implement `Auditor` on `PhenotypeCreator`.

The phenotype creator checks some basic HPO term issues:
- the input `str` is a valid CURIE
- the CURIE corresponds to an HPO term id
- the term id is in the used ontology
- the term id is not obsolete

The offenders are dropped or updated (obsolete term IDs).

The rest is delegated to HPO toolkit's `ValidationRunner`, which should complain about terms not being descendants of *Phenotypic abnoramlity*, and about annotation propagation violations.


ielis:
@lnrekerle please have a look, please check the notebooks as we discussed, and give it a :+1: if looks good.

lnrekerle (approved):
Sorry this took me a bit to get to, but looks good!

Issue #102, opened 2023-11-21
`ProtFeatureTypePredicate` and `ProtFeaturePredicate` should use regions to determine overlaps


ielis:
The protein predicates should use functionality of `Region` (introduced in #101 ) to determine the overlap instead of the hard-to-read statements like:

![Image](https://github.com/monarch-initiative/genophenocorr/assets/12170955/fadbbc9c-0c45-4c69-95dd-8b47e8292b18)
> L#186

or:

![Image](https://github.com/monarch-initiative/genophenocorr/assets/12170955/b1add345-cf88-466a-b24f-6df960df45b1)

> L#229

PR #130, opened 2024-02-12
Protein Service

### LINKED ISSUE #118: Remove `ProteinMetadata` from `TranscriptAnnotation`
As of now, `TranscriptAnnotation` includes `protein_affected` property with a sequence of `ProteinMetadata`. However, the property is out of scope and should be removed.

Instead, we just need to keep the `protein_effect_location` that describes the location of the variant within the protein sequence, and we must add a new property with `protein_id` to ensure we can map `tx` to the corresponding protein accession.

## Requirements

Removing `ProteinMetadata` propagates across our code base and it will require multiple updates, such as the following:

- Remove `protein_affected` property from `TranscriptAnnotation`
- add new property with a good name that will return the protein accession ID (e.g. `NP_037407.4` for `NM_013275.6` [here](https://github.com/monarch-initiative/genophenocorr/blob/3082135b701182b1ef51947335d70ac37856261c/src/genophenocorr/preprocessing/test_data/vep_response/deletion.json#L69)). The property provides an `Optional[str]` because some transcripts are not protein-coding
- adjust `VepFunctionalAnnotator`. 1) remove `protein_annotator` because we do not need it anymore, right? 2) implement parsing of the new property with protein ID
- update `Cohort.get_protein_features_affected` - when we don't have `ProteinMetadata` in `Variant`s, then we won't be able to produce this information anymore. I think this is not too bad and we should just report the protein IDs.
- update `Patient` and `PatientCreator` - the `Patient` model should not be aware of the `ProteinMetadata`. The model should just expose the variants and phenotypes - the raw data from phenopackets. Therefore, `proteins` property must be removed. In result, this propagates into the `Cohort.from_patients` method. We will need to rework the logic there too.

I suggest we should do these changes together during a call, during a "stop the world" session, when much else is going on in the code, PRs are merged and these changes will not break stuff on other branches.

lnrekerle:
Closes #118 and #102 

Moved Protein Metadata Service out of the original set up and into the analysis, so that we do not reach out to Uniprot for unnecessary proteins. 

Also, changed each protein analysis to use the Region function "overlaps_with" rather than the long test before.


ielis (changes requested):
Hi @lnrekerle 
The CI must pass, that's the first requirement.

src/genophenocorr/analysis/_config.py
```diff
@@ -165,20 +168,47 @@ def build(self) -> CohortAnalysisConfiguration:
...
                               hpo: hpotk.MinimalOntology,
+                              protein_source: str = 'UNIPROT',
+                              cache_dir: typing.Optional[str] = None,
                               config: typing.Optional[CohortAnalysisConfiguration] = None) -> CohortAnalysis:
     """
     Configure :class:`genophenocorr.analysis.CohortAnalysis` for given `cohort`.
 
     :param cohort: a :class:`genophenocorr.model.Cohort` to analyze
     :param hpo: a :class:`hpotk.MinimalOntology` with HPO to use in the analysis
+    :param protein_source: the resource to retreive protein annotations from if we cannot find the annotations locally.
```
ielis, line 179: A typo `retrieve`..

src/genophenocorr/model/_cohort.py
```diff
@@ -299,6 +276,7 @@ def get_excluded_ids(self):
     def get_excluded_count(self):
         return len(self.all_excluded_patients)
 
+    """
```
ielis, line 279: The block is commented out in a very creative way!

It would be probably better to use `#` to comment out the function, if, for some reason, you prefer to keep the code around.

src/genophenocorr/model/_variant.py
```diff
@@ -377,13 +379,13 @@ def create_variant_from_scratch(variant_coordinates: VariantCoordinates,
                                     is_preferred: bool,
                                     consequences: typing.Iterable[VariantEffect],
                                     exons_effected: typing.Sequence[int],
-                                    protein: typing.Sequence[ProteinMetadata],
+                                    protein_id: typing.Sequence[str],
```
ielis, line 382: This should probably be `typing.Optional[str]` instead of `typing.Sequence[str]`, since an optional `str` is what `TranscriptAnnotation` expects.

src/genophenocorr/preprocessing/_phenopacket.py
```diff
@@ -162,7 +162,7 @@ def process(self, inputs: Phenopacket, notepad: Notepad) -> Patient:
             Patient: A Patient object
         """
         sample_id = SampleLabels(label=inputs.subject.id, meta_label=inputs.id if len(inputs.id) > 0 else None)
-
+        #print(f"Working on sample {sample_id}")
```
ielis, line 165: Pls remove :)

tests/fixtures.py
ielis, line 1: I read a bit about Pytest fixtures, and I think we should rename this file to `conftest.py`. Thanks to renaming, pytest will pick up the fixtures automatically, without having to import them as we do e.g. on `test_predicates.py` L#11. 

`conftest.py` is a special file where pytest looks for fixtures.

tests/fixtures.py
```diff
@@ -1,106 +1,88 @@
...
 
+from genophenocorr.data import ProteinTestMetadataService
 from genophenocorr.model import *
 from genophenocorr.model.genome import GRCh38, GenomicRegion, Region, Strand
 
-
 def make_region(contig: str, start: int, end: int) -> GenomicRegion:
     return GenomicRegion(GRCh38.contig_by_name(contig), start, end, Strand.POSITIVE)
 
+@pytest.fixture
```
ielis, line 11: I think we can set `@pytest.fixture(scope='session')` here. A fixture with `scope='session'` is created only once and used for _all_ tests, as opposed to creating a fixture for each test.

A small detail, but perhaps worth knowing about.

src/genophenocorr/data/_protein_test_service.py
ielis, line 1: I think this file belongs to `tests/` subhierarchy and not into `src/genophenocorr` (production).

I did not find any use of `ProteinTestMetadataService` (essentially a mock object) in the production code, we seem to be using it only in tests. So, unless I am missing something, moving to `tests/` is appropriate.

ielis (approved):
Hi @lnrekerle 

thank you for the PR! 

I pointed out a few cosmetic details that I'd like to ask you to consider. However, these issues are _not_ mission critical. Otherwise, the code is ready to go.

tests/fixtures.py
lnrekerle: We already have a conftest.py file. should I combine the two? It looks doable, there is not a lot in the conftest file.

ielis: Yes, I think combining should be done. The purpose of `fixtures.py` is to define fixtures. However, in `pytest`, `conftest.py` is the right/idiomatic way for defining fixtures.

I learned about this only recently, but pls merge the files such that we have `conftest.py` with the global fixtures. Note that in general, we can have `conftest.py` in each test module, which can come in handy in certain use cases.

PR #138, opened 2024-02-22
changed all occurances of hgvsc_id to hgvs_cdna

### LINKED ISSUE #111: TranscriptAnnotation hgvsc_id
The field hgvsc_id in TranscriptAnnotation seems to have values such as 
```
'NM_001330437.2:c.836A>G'
```

This is not the id it is the actual HGVS cDNA notation. If this is meant to be used as an id, then we should not put the actual mutation into the id, it is confusing. I am assuming that the variable name is wrong and it should be changed to 

hgvs_cdna (also use hgvs_protein)

lnrekerle:
Closes #111 

Changed all occurrences of hgvs_id to hgvs_cdna to make the variable name more clear.


ielis (approved):
Hi @lnrekerle 
there is indeed not much more to this issue.

PR #143, opened 2024-03-07
Add Disease to Patient

### LINKED ISSUE #140: Extract disease
Extract disease from phenopacket
- Disease is represented in two places. Please extract from the diseases list of the top level (https://phenopacket-schema.readthedocs.io/en/latest/phenopacket.html)
- Extract more than one if multiple diseases present
- If no disease found, WARN (previous phenopackets won't work)
- Write test case to check for correct diagnosis from importing a phenopacket
- You can use phenopackets from https://github.com/monarch-initiative/phenopacket-store/tree/main/notebooks/11q_terminal_deletion/phenopackets. (fine to shorten the phenopapcket to have only 3 HPOs). Check for disease id and label. If there is a disease element without both, this is an error

lnrekerle:
Closes #140 

This created the class Disease and the function _add_disease. 

_add_disease takes a list of Phenopacket Disease objects and returns a list of this new Disease class objects.


pnrobinson (approved):
+1

PR #148, opened 2024-04-23
Draw variants transcript and protein

pnrobinson:


pnrobinson:
@ielis @lnrekerle 
This PR adds the drawing code as well as a test case for the VariantValidator response. The function, parse_response, line 157 of _vv.py, will also need to be changed -- to me it looks as if the code is expecting a different JSON structure from what we actually get -- I added the test case to understand the structure. Unless there is some other code that only passes part of the response, it is wrong
PLEASE ADD DOCUMENTATION TO ALL FUNCTIONS LIKE THIS! It would be a lot easier to contribute code if the main assumptions and goals of each function were documented --

lnrekerle (approved):
This looks good to me!

lnrekerle:
> The function, parse_response, line 157 of _vv.py, will also need to be changed -- to me it looks as if the code is expecting a different JSON structure from what we actually get -- I added the test case to understand the structure. Unless there is some other code that only passes part of the response, it is wrong PLEASE ADD DOCUMENTATION TO ALL FUNCTIONS LIKE THIS! It would be a lot easier to contribute code if the main assumptions and goals of each function were documented --

That function was originally written by @ielis, but I can take a look at it and see if I can recreate and fix the bug. I do think that this PR as it is is ready to be pushed.

pnrobinson:
I am going to merge this, and now the code seems to be working again. I am not changing _vv.py now but we may need to do so, but let's make that a separate PR

PR #154, opened 2024-05-08
Kbg

pnrobinson:
This PR adds a Jinja2 template for displaying basic data about a cohort.
I have also updated the KBG cohort and renamed it according to the gene (ANKRD11) -- I think it is best if we use gene names for our cohorts and folders for consistency. 
See the notebooks/ANKRD11 folder for an example of using the new template.
I think that we still need to do something to make sure that the Python package will include the template in the distribution -- I have not worked on this yet, but it should be fine for testing from the Git folder.


pnrobinson:
I see the CI is not finding the Jinja package
```
No matching distribution found for Jinja>=3.1.4
```
This is available via PyPI so I am not sure

ielis:
It [looks](https://github.com/pallets/jinja/blob/a516a99bab2f401771964e5c42f09bcf57492e41/pyproject.toml#L2) like the package name is actually `Jinja2`.

Let me try to fix this.

ielis (approved):
@pnrobinson I updated the class for summarizing the cohort. I think `PackageLoader` is better for loading the templates since we ship them in the package. 

> I think that we still need to do something to make sure that the Python package will include the template in the distribution -- I have not worked on this yet, but it should be fine for testing from the Git folder.

I included the template folder in the manifest to ensure the templates will be part of the package.


Next, I updated the notebooks. Basically, we should use the summary template everywhere after loading the cohort, to summarize the loading process. 

In overall, I like the idea of working with Jinja templates, I think we should use the templates to report the MTC results as well. So, the template is OK as well as the PR. What should the next steps be?

pnrobinson:
@ielis 
OK, I will merge this for now
"What should the next steps be?" -- The most important thing would be to get the multiple testing working, if there is time for a hacking session to get started that would be great. We then need to polish the graphics and to add the Wilcoxon test

PR #155, opened 2024-05-08
Buffer vv response

pnrobinson:
I made some changes in the variant drawing code because of None-pointer errors, e.g., 
```
     variant_locations = list()
      for ann in tx_anns:
          if ann is not None and hasattr(ann, 'protein_effect_location'):
              prot_eff_loc = ann.protein_effect_location
              if prot_eff_loc is not None:
                  variant_locations.append([prot_eff_loc.start, prot_eff_loc.end])
      variant_locations = np.array(variant_locations)
      #variant_locations = np.array([[
      #    ann.protein_effect_location.start,
      #    ann.protein_effect_location.end]
      #    for ann in tx_anns if ann is not None and hasattr(ann, 'protein_effect_location')
      #])
```
This solved that problem, but I am seeing strange behaviour
Screenshot 2024-05-08 at 16 02 03

I am not sure if the problem is related.


pnrobinson:
crap, I think this PR included changes from another PR -- it is really just the change in the code that is above comment, please delete this PR and just adopt the change if it is correct

ielis:
> crap, I think this PR included changes from another PR -- it is really just the change in the code that is above comment, please delete this PR and just adopt the change if it is correct

yes, it looks like the changes from another PR include deleting some phenopackets. I see 328 *KBG* phenopackets gone on my diff. I think we should wait until `KBG` PR is merged and then we should be able to merge this one.

src/genophenocorr/preprocessing/_vv.py
```diff
@@ -155,17 +157,24 @@ def _parse_tx(tx: typing.Union[str, TranscriptInfoAware]) -> str:
...
-        if len(response) != 1:
-            self._logger.warning('Response has %s!=1 items. Choosing the first', len(response))
-
-        item = response[0]
-        if 'requested_symbol' not in item or item['requested_symbol'] != tx_id:
-            raise ValueError(
-                f'Could not find {tx_id} in the `requested_symbol` field in the response from Variant Validator API'
-            )
-        if 'transcripts' not in item:
+        if not isinstance(response, list):
```
ielis, line 160: The aim of this function is to process the response into `TranscriptCoordinates` or to raise a `ValueError` if there are any issues with the input.

I wrote few tests for you to play with the function. Please figure out what you are after and remove the print statements at the end so that we do not litter STDERR :)

ielis (changes requested):
@pnrobinson please see the comment below.

src/genophenocorr/preprocessing/_vv.py
pnrobinson: @ielis I should have left this as draft. The problem is that response from VariantValidator is not always exactly what this function expects, and I I added the print statement to get better feedback than from throwing an error. Testing is hard because we do not have all inputs but I will try to get something reasonable

ielis: Hm.. :/ If you're seeing errors, then test suite cannot be comprehensive, I only tested ~4 transcripts :)

Please let me know when you receive a weird response from VV, I'll update the tests & fix the code.

pnrobinson: Working with this, I think the PR is ready to be merged so we can move ahead. I cannot reproduce the bud I was seeing anymore.

pnrobinson (commented):
The print statements are removed, I will merge now as I think this is what you are requesting and it will simplify the rest of the work.

PR #160, opened 2024-05-18
Mtc

pnrobinson:
This PR implements the heuristic method for choosing which terms not to apply tests to. On an example, it skips over 50 terms for correct reasons and this helps with the MTC burden.


pnrobinson:
@ielis @lnrekerle 
see this notebook https://github.com/monarch-initiative/genophenocorr/blob/mtc/notebooks/MAPK8IP3/MAPK8IP3.ipynb
for an example of the MTC strategy in action -- there is a table showing that we omit over 100 terms prior to testing.

lnrekerle (approved):

pnrobinson:
merging to simplify development in next few days.

PR #162, opened 2024-05-19
protein drawing

pnrobinson:
This PR adapts the code of @frehburg and outputs just a protein drawing with the correct positions.
It needs a little TLC but it seems to work

TLC includes
1. Get the protein ID automatically from the transcript annotation (I was using the wrong one and trying to catch the bug for 30 min, but the variant validator provides the protein id)
2. Check whether the ProteinVisualizable is valid before plotting it and of not output an error message rather than plotting "whatever"
3. Get the protein length from the uniprot API (right now I am taking the maximum length of the features or variants, but this is wrong)
4. Get the name of the protein from the uniprot API or from variant validator (name of gene) or both
5. Clean up the ProteinVisualizer object to return exactly what the visualizer needs.


lnrekerle (approved):

pnrobinson:
Merging to have everything in same branch for testing examples.

PR #167, opened 2024-05-20
Added disease viewable, variant and effect counts

lnrekerle:
Created a viewable for diseases that lists how many of each variant/variant effect is represented in the cohort by disease.


pnrobinson (approved):
+1

PR #169, opened 2024-05-21
Kmt2 d

pnrobinson:


src/genophenocorr/view/_protein_visualizer.py
```diff
@@ -146,29 +161,22 @@ def draw_fig(self, pvis:ProteinVisualizable):
...
-            print(f"type max_x_absolute {type(max_x_absolute)} ")
-            if x_absolute.any() > max_x_absolute:
-                print(f"x={x_absolute} but max={max_x_absolute}")
-            shifted_to_0_1 = ((x_absolute - min_x_absolute) / (max_x_absolute - min_x_absolute))
-            relative_scale = (protein_track_x_max - protein_track_x_min)
-            return shifted_to_0_1 * relative_scale + protein_track_x_min
+        def preprocess(absolute,
+                       min_absolute=min_aa_pos, max_absolute=max_aa_pos,
+                       min_relative=protein_track_x_min, max_relative=protein_track_x_max):
+            if absolute.any() > max_absolute:
```
ielis, line 174: The context suggests that `absolute` is a numpy array with `float`s or `int`s. What is the meaning of calling `.any()` on such array?

ielis: But then it should be:
```python
(absolute > max_absolute).any()
```

I'll fix it, thanks a lot for the info.

ielis: With respect to the 5 or 3UTR variants, we should we put them at the start (`0`) or the end (`len(protein)`)  coordinates? Basically do the same thing we intended for intronic variants, putting them close to the place where they are relevant?

Please shout if this is a bad idea, otherwise I will place the variants there, so that we show *all* variants in the figure.

ielis (approved):
Looks great, thank you!

PR #173, opened 2024-06-11
Notebooks all work. Small changes so tables look good.

lnrekerle:
This PR adds a variant name column that shows the HGVS name of a variant in addition to the chromosomal representation. Currently the PR only works for small variants (HGVS). We will additionally need a function for structural variants. 
The tutorial.rst was also updated for the change in the variant_effect_count_by_tx function. 
And finally, the function get_hgvs_cdna_by_tx was added to the Class Variant.


ielis:
Hi @lnrekerle pls request review when this is ready to be looked at and when the required checks pass. Thanks!

pnrobinson:
@lnrekerle 
Please check why the CI test is failing. It seems there is one item in the documentation that is not up to date
```
tests/analysis/test_mtc_filter.py: 740 warnings
  /home/runner/work/genophenocorr/genophenocorr/src/genophenocorr/analysis/predicate/phenotype/_pheno.py:153: DeprecationWarning: `is_observed` property was deprecated and will be removed in `v0.3.0`. Use `is_present` instead
    if phenotype.is_observed:
````
In general we want all tests to pass before we merge a PR, let's try to clean everything up!

**Comment by pnrobinson** [comment]
@lnrekerle 
Please also write a brief summary of what the PR does.
From our discussion yesterday, I think it is the addition of the "Variant name" column. This is nice, and I looked at some of the notebooks and it seems to work.
So for instance, please write something like this

"This PR adds a variant name column that shows the HGVS name of a variant in addition to the chromosomal representation."
Probably also: "Currently the PR only works for small variants (HGVS). We will additionally need a function for structural variants" (this lets the reviewer know exactly what the PR does!)

**Review by pnrobinson — APPROVED** [review]
OK, looks good!

File: src/genophenocorr/model/_cohort.py [hunk 1]
```diff
@@ -273,15 +273,15 @@ def variant_effect_count_by_tx(
             mapping: Each transcript ID references a Counter(), with the variant effect as the key
               and the count of variants with that effect on the transcript id.
         """
-        counters = defaultdict(Counter)
+        #counters = defaultdict(Counter)
```
ielis, line 276: Comments like this serve no purpose.

src/genophenocorr/model/_variant.py
```diff
@@ -423,6 +423,20 @@ def tx_annotations(self) -> typing.Sequence[TranscriptAnnotation]:
     @property
     def genotypes(self) -> Genotypes:
         return self._gts
+    
+    def get_hgvs_cdna_by_tx(self, transcript_id:str) -> str:
```
ielis, line 427: Please verify if the typing annotation for the return type (and the documentation) is correct.

ielis (changes requested):
Hi @lnrekerle

please review my comments below and verify that the notebook at `notebooks/ANKRD11/KBG.ipynb` is al right.

src/genophenocorr/model/_variant.py
lnrekerle: updated documentation and set return type to Optional[str]

src/genophenocorr/model/_cohort.py
lnrekerle: Removed comment

pnrobinson (approved):
@lnrekerle  Please fix all of the errors in the notebooks, e.g.  in KBG

cannot import name 'StatsViewable' from 'genophenocorr.view' (/Users/rekerl/Documents/Programs/genophenocorr/src/genophenocorr/view/__init__.py)

Our goal should be that the example notebooks always are up to date and work, and if we change something then the notebooks should be revised as needed. After this I think we should merge this request

lnrekerle (commented):
I am unable to merge this into develop until changes have been addressed, so testing to see if I "review changes", if that will let me merge. Otherwise I might need Daniel to confirm that I did address his change requests. 

Edit to add: this did not work, I am still unable to merge this PR myself. @ielis & @pnrobinson

src/genophenocorr/preprocessing/_protein.py
```diff
@@ -35,7 +35,10 @@ def get_annotations(self, protein_id: str) -> typing.Optional[ProteinMetadata]:
         fpath = self._create_file_name(protein_id)
         if os.path.isfile(fpath):
             with open(fpath, 'rb') as fh:
-                return pickle.load(fh)
+                result_list = pickle.load(fh)
```
ielis, line 38: I think you may be fixing the wrong thing here.

The change of `ProteinMetadataService` in `cd6dfa295e07d878d9b71e7ff1498056a7b51a3c` results in returning (and pickling) of a singluar instance, and not a list. Therefore, the pickle files are obsolete and must be updated. The code of this function should not be changed.

In other words, previously, we may have been serializing a list of `ProteinMeatadata` instead of a singular instance. Therefore, you have weird old pickle files, and it seems to me you want to fix them by updating the code, while the solution is to wipe the old pickle files.

ielis (changes requested):
@lnrekerle, sadly, this is still wrong and must be fixed. I added few tests to prove that the Python documentation strings of the two methods that are updated by this PR are lies.

Please make sure the tests pass.

src/genophenocorr/preprocessing/_protein.py
lnrekerle: Yah, I reran this after deleting the .genophenocorr_cache/ folder and it ran fine before the changes. I didn't realize how old my files were. this is fixed.

ielis:
@lnrekerle the merge honors are all yours!

PR #177, opened 2024-06-28
Adding new MTC filter that allows the user to specific which term(s) …

pnrobinson:
…they want to test. Only these terms are tested. Also added new unit test in test_mtc_filter.py


pnrobinson:
I also tested this assertion (https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8948816/  "The frequency of global developmental delays and intellectual disability/learning difficulties in patients with truncated ANKRD11 gene variation was higher than that in patients with missense variation in the ANKRD11 gene (p < 0.05).", made on only 13 patients) by testing only two terms in the KBG notebook. There was no significance, and so I think we can say that we have fairly ruled this out.

ielis (approved):
I moved the validation of the specified HPO terms to the configuration to catch weird HPO terms and "explode" as early  in the analysis as possible, as a courtesy to our users.

Otherwise, it looks great, thanks a lot!

pnrobinson:
@ielis thanks! I will merge

PR #178, opened 2024-07-01
More mtc heuristic

pnrobinson:
I could use a hand in setting up a unit test -- can we make a small cohort with two HPO terms and missense vis nonsense?


ielis:
Maybe we could use a Phenopacket store cohort? Would you have a candidate there?

pnrobinson:
This PR adds the ability to set a minimum frequency to have in at least one genotype for the HeuristicMtcFilter
See test_min_observed_HPO_threshold in test_mtc_filter for the key part of the function.

ielis (approved):
Looks good, thanks!

PR #179, opened 2024-07-01
Create Protein table so users can see list of protein features and coordinates

### LINKED ISSUE #176: Number Protein Regions
Update ProteinMetadata so that each Feature has a unique label. 
i.e. - if there are multiple Features labeled "Disordered", they would become "Disordered_1", "Disordered_2", etc. 

Definition of Done (DoD): A unit test is created that has an example protein with multiple Features of the same name, an analysis passes and gets expected output for just one of the Features and it does not test with all of them. 

EDIT: After discussion, we will not be numbering features and will instead list them in a table for users to select. 

New DoD: A proteins table is printed with viewer that lists all protein features and the location on the amino acid strand.

lnrekerle:
Rather than having multiple duplicate feature names, if multiple are named "Disordered" they will be relabeled as "Disordered_1", "Disordered_2", etc. This will make it easier for a user to pick a specific area when running the "compare_by_protein_feature" predicate. 

I also added an assertion to verify that they are counted properly. 

This will close #176 

EDIT: Updated to reflect updates in #176 
This PR creates a class called ProteinViewable that prints a table that shows the protein features and coordinates.


pnrobinson (approved):
+1

lnrekerle:
Screenshot 2024-07-09 at 10 46 56 AM
Example!

src/genophenocorr/view/_protein_viewer.py
```diff
@@ -0,0 +1,48 @@
...
+    @staticmethod
+    def _get_start(feat_list: typing.Iterable[typing.Any]) -> int:
+        return feat_list[2].start
+    
+    def _prepare_context(self, cohort: Cohort, pvis: ProteinVisualizable) -> typing.Mapping[str, typing.Any]:
+        protein_id = pvis.protein_id
+        protein_features = []
+        
+        for i in range(len(pvis.protein_feature_names)):
+            protein_features.append([pvis.protein_feature_names[i], pvis._protein_feature_types[i], Region(pvis.protein_feature_starts[i],pvis.protein_feature_ends[i]), 0])
```
ielis, line 32: Right now, `protein_features` is a list of lists and it takes a significant amount of brain power to understand what does it do.

Can you please make `protein_features` a list of named tuples? If you do not know what a named tuple is, please look it up in Python's documentation. Then create a named tuple that is private to this module and make a tuple for each list item. Then use the named tuple's properties in the downstream code.

The code is probably correct, but with named tuples you will make it easier to maintain and linters (and Python interpreter FWIW) will stand by you and help you find mistakes.

src/genophenocorr/view/_protein_visualizable.py
```diff
@@ -1,4 +1,5 @@
 import typing
+import regex as re
```
ielis, line 2: Why do we need this `regex` library? Is it better than Python's builtin `re` package?

src/genophenocorr/view/templates/protein.html
ielis, line 1: When looking at the example in the PR, I find the colums a bit too close together. It would probably be good to add a bit of vertical padding between the columns (optional).

pyproject.toml
```diff
@@ -42,7 +42,8 @@ dependencies = [
     "statsmodels>=0.13.0",
     "numpy>=1.23",
     "matplotlib>=3.2.0,<4.0",
-    "tqdm>=4.60"
+    "tqdm>=4.60",
+    "regex>=2024.5.15"
```
ielis, line 46: Why do we need `regex`? Is Python's builtin `re` insatisfactory? An honest question..

ielis (changes requested):

src/genophenocorr/view/_protein_visualizable.py
lnrekerle: I actually removed what was using regex, so neither are needed anymore. 
But to answer your question, I forgot that Python had it a built in

ielis: Python is "batteries included" for many things.. :)

src/genophenocorr/model/_variant.py
```diff
@@ -424,6 +424,21 @@ def tx_annotations(self) -> typing.Sequence[TranscriptAnnotation]:
     def genotypes(self) -> Genotypes:
         return self._gts
     
+    def get_tx_anno_by_tx_id(self, transcript_id:str) -> typing.Optional[TranscriptAnnotation]:
```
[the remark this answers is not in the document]
lnrekerle: Moved this to FunctionalAnnotationAware!

src/genophenocorr/model/genome/_genome.py
```diff
@@ -240,6 +240,12 @@ def end(self) -> int:
         Get 0-based (included) end coordinate of the region.
         """
         return self._end
+    
+    def to_string(self) -> str:
```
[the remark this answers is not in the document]
lnrekerle: I decided to remove it.

ielis (approved):
Looks good now, thanks!

I found that an "unfrozen" dataclass is a bit better than `namedtuple`. Named tuples are immutable so they do not fit this use case very well.

Please merge at your convenience!

PR #182, opened 2024-07-04
fixing template to have blocks appear underneath each other

pnrobinson:
This PR fixes the CSS so that the three tables appear one under the other, rather than three in an uneven row as before. I will merge this soon unless somebody protests, it is a very local change.


ielis (approved):
Looks great, thank you!

PR #184, opened 2024-07-09
Merge Protein Visualizer into develop

frehburg:
Updates:
- figure can display overlapping domains
- improved code readability and structure
- legend for variants added
- readable protein feature domain labels
- deterministic & reproducible coloring


pnrobinson:
@frehburg 
The CI is reporting this error:
```
src/genophenocorr/view/_protein_visualizer.py:154: in <module>
    @dataclass(slots=True)
E   TypeError: dataclass() got an unexpected keyword argument 'slots'
```
This is strange because you showed me the code working fine today. Do you know what is going on? In general, we never accept a PR if the CI is broken! Thanks!

frehburg:
Dear Peter, oh I am sorry. The feature `slots` was only added in Python 3.10 if I remember correctly. I can revise the code while keeping it's memory efficiency tomorrow.

https://docs.python.org/3/library/dataclasses.html

pnrobinson:
If this is a useful feature, I do not see much point in supporting many backward versions of Python -- @ielis what about supporting Python 3.10 and upwards? We are at 3.12 now and could also add that to the CI?

ielis:
I looked at the [docs](https://docs.python.org/3/library/dataclasses.html) and I do not see any benefits of using `slots=True` at the cost of removing support for Python <3.10. @frehburg are there any strong arguments for using this option?
IMO, support of broader Python versions is more useful than anything else I can see at this point, because an older Python may be all that people have in some constrained environments.

So, unless there are *strong* arguments for using `slots=True`, we should probably avoid it.
@frehburg please let us know.

frehburg:
Dear @ielis,

as I said. I will rewrite that bit of code today. There are other options.

Regards,

Filip

frehburg:
@pnrobinson @ielis There is a way to manually define slots, which was already available in 3.8. It's a bit less neat, but preserves the compatibility with older Python versions.

I hope this works.

ielis (approved):
I didn't really check this line by line but it looks reasonable, and I think we will be able to troubleshoot the code in future, if necessary. Thanks a lot!

PR #185, opened 2024-07-19
Created a Formatter

### LINKED ISSUE #109: "toString" for Variant
It would be nice to be able to do
```
 print(f"[WARNING] Could not find annotation of variant {v} for {reference_tx_id}")
```
instead of (as now)
```
print(f"[WARNING] Could not find annotation of variant {v.variant_coordinates.variant_key} for {reference_tx_id}")
```
to get
```
[WARNING] Could not find annotation of variant 12_112473023_112473023_A_G for NM_002834.5
```

lnrekerle:
Created a class called Formatter, which is a metaclass that can then be used across many different classes in genophenocorr/model to make them easier to read when being printed in tables or figures. We could eventually add more formatters, or remove the metaclass and just use VariantFormatter if it seems unnecessary. 

Parts of this may need to change or be added once we add SV, but so far, every SV I could find would have at least a variant key, if not also an hgvs string. 

This will close #109


lnrekerle:
Oops! Definitely meant to add you both as reviewers! @ielis @pnrobinson 
Thanks!

ielis (changes requested):
The code looks OK, there are some good things but also some things that need change. Please see my comments below and please make sure to ask for re-review when you do the updates.
Thank you!

ielis (changes requested):
Hello, few more things to be addressed are listed below.

ielis (approved):
Looks good, thanks!

PR #189, opened 2024-07-23
multiple testing documentation

pnrobinson:
This adds documentation about how to use the MTC functions. We will need something similar for each of the main functionalities of genophenocorr.


pnrobinson:
I am getting this "The job was canceled because "_3_11" failed.", but will this be fixed by another PR?

ielis (dismissed):
The CI fails because it involves reaching out to VEP to annotate some variants, and the request times out. However, the tests pass locally on my side. It is a bit annoying, but we will need to address this another time.

Otherwise, this PR is excellent. I simplified the MTC user guide to only deal with the `CohortAnalysisConfiguration`, and omit all the other imports, because I think they are irrelevant. We do not really need to run an *entire* analysis in the MTC guide, because we do that elsewhere. We only show how to use this or that option for tweaking the desired MTC parameters.

@pnrobinson please have a look at the updated text (if you'd like to). Otherwise, I think this is great and ready to merge.

Therefore, I'm changing it to PR, giving my thumbs up, and we can force-merge despite CI failures.

ielis (approved):
Awesome!

PR #193, opened 2024-07-26
Removed the red background 

lnrekerle:
Removed the red background when running load_phenopacket_folder by taking the notebook warnings out of the stderr and putting them in stdout. I noticed everything was printing to the stderr file, and that always prints as red in jupyter-notebook. By moving the warnings out of the stderr file, only the errors will return with a red background now. 

Also discovered tqdm has a module specifically for jupyter-notebook, so I added that. I like that the progress bar is now blue/green ☺️

### DIFF
--- a/src/genophenocorr/preprocessing/_config.py
+++ b/src/genophenocorr/preprocessing/_config.py
@@ -287,12 +287,12 @@ def _summarize_validation(policy: str,
                     for warning in node.warnings():
                         lines.append(l_pad + ' ·' + warning.message
                                      + (f'. {warning.solution}' if warning.solution else ''))
-        sys.stderr.write(os.linesep.join(lines))
+        print(os.linesep.join(lines), file=sys.stderr)
     else:
         lines.append('No errors or warnings were found')
         l_pad = ' ' * (notepad.level * indent)
         lines.append(l_pad + notepad.label)
-        sys.stdout.write(os.linesep.join(lines))
+        print(os.linesep.join(lines))
 
 
 def _load_phenopacket_dir(pp_dir: str) -> typing.Sequence[Phenopacket]:
--- a/src/genophenocorr/preprocessing/_config.py
+++ b/src/genophenocorr/preprocessing/_config.py
@@ -7,7 +7,7 @@ import hpotk
 # pyright: reportGeneralTypeIssues=false
 from google.protobuf.json_format import Parse
 from phenopackets import Phenopacket
-from tqdm import tqdm
+from tqdm.notebook import tqdm
 
 from genophenocorr.model import Cohort
 from genophenocorr.model.genome import GRCh37, GRCh38, GenomeBuild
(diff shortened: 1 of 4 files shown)


src/genophenocorr/preprocessing/_config.py
```diff
@@ -241,13 +241,15 @@ def load_phenopacket_folder(
...
     # Keep track of the progress by wrapping the list of phenopackets
     # with TQDM 😎
-    cohort_iter = tqdm(pps, desc='Patients Created')
+    cohort_iter = tqdm(pps, desc='Patients Created', file=sys.stdout, colour='green')
     notepad = cohort_creator.prepare_notepad(f'{len(pps)} phenopacket(s) found at `{pp_directory}`')
     cohort = cohort_creator.process(cohort_iter, notepad)
-
-
+    
     validation_summary = _summarize_validation(validation_policy, notepad)
```
ielis, line 248: This is OK in principle, but I think it would be better to report *both* errors and warnings to STDERR (red), and report `No errors or warnings were found` to STDOUT (no red).
To do this, we would either need to return more data from `_summarize_validation`, or to deal with reporting the results within `_summarize_validation`. I think we should do the latter.

To make this happen, let's return nothing from `_summarize_validation` and deal with printing to streams within the function body. More info follows below.

src/genophenocorr/preprocessing/_config.py
```diff
@@ -267,30 +269,33 @@ def load_phenopacket_folder(
 
 def _summarize_validation(policy: str,
                           notepad: NotepadTree,
-                          indent: int = 2):
-    lines = [f'Validated under {policy} policy']
+                          indent: int = 2) -> typing.Tuple[typing.Sequence[str], typing.Sequence[str]]:
```
ielis, line 272: Let's not return anything here.

src/genophenocorr/preprocessing/_config.py
```diff
@@ -267,30 +269,33 @@ def load_phenopacket_folder(
...
+    warn_lines = [f'Validated under {policy} policy']
+    error_lines = []
     n_errors = sum(node.error_count() for node in notepad.iterate_nodes())
     n_warnings = sum(node.warning_count() for node in notepad.iterate_nodes())
     if n_errors > 0 or n_warnings > 0:
-        lines.append('Showing errors and warnings')
+        warn_lines.append('Showing errors and warnings')
```
ielis, line 278: Let's only print in RED if we find any warnings or errors and we follow this branch.

We can collect the `warn_lines` and `error_lines` in this branch and report the lines to STDERR (red) at the end of the branch body.

src/genophenocorr/preprocessing/_config.py
```diff
@@ -267,30 +269,33 @@ def load_phenopacket_folder(
...
-                    lines.append(l_pad + ' warnings:')
+                    warn_lines.append(l_pad + ' warnings:')
                     for warning in node.warnings():
-                        lines.append(l_pad + ' ·' + warning.message
+                        warn_lines.append(l_pad + ' ·' + warning.message
                                      + (f'. {warning.solution}' if warning.solution else ''))
     else:
-        lines.append('No errors or warnings were found')
-    return lines
+        warn_lines.append('No errors or warnings were found')
```
ielis, line 295: Here we found no errors. We can print the corresponding text to STDOUT (no red).

ielis (changes requested):
Hi @lnrekerle,

In principle this looks good. However, I think now the code reports warnings to STDOUT and errors to STDERR. However, I think we should report *both* warnings and errors to STDERR and print to STDOUT only if no errors or warnings are found.

I am suggesting how to do this in the comments below. Otherwise this looks good.

ielis (approved):
Thanks a lot, looks great now!

ielis:
@lnrekerle I think this is good despite CI failures. The failures are unrelated, due to REST API timing out. We'll have to do something with it at some point. However, I think this is ready to go and, therefore, I will merge.

PR #197, opened 2024-07-31
Add `CountingPhenotypePredicate`

ielis:
Add `CountingPhenotypePredicate` to bin patients according to the number of broad phenotypic features.


pnrobinson:
@frehburg sorry I forgot that you were maybe working on this? Can we collab on the unit tests if you are around? 
@ielis 
can you do a quick sanity check on CountingPhenotypePredicate -- if this is OK, we will do the unit tests

ielis:
The unit tests fail, but this is heading the right direction. Please make the tests pass, then we should be all set.

pnrobinson:
The PhenotypePredicate object is not really relevant for a test that is not based on a 2x2 or a 2x3 contingency table.
I had to write a class called ``IntegerCountCategory'' to get the API to not complain. For the tests, I had to do this
```
PhenotypeCategorization(IntegerCountCategory(0), 0)
```
This is not particularly elegant, but I do not think that the statistical test (Mann Whitney U test) needs to have the categories, and so it might be ok just to keep things as is. @ielis could you sanity check. If this is OK, then I will write the stats test on Monday.

ielis (commented):
We don't need `IntegerCountCategory`, `PatientCategory` will do OK.

I think the code is now OK, let's move ahead to writing the stats test.

PR #198, opened 2024-08-01
Add section about statistical tests into the documentation

pnrobinson:
Hi!
Here is an example of how to use the Mann Whitney test (see new stats.rst page).
Unfortunately, this test is not implemented in statsmodels, and I could find an implementation only in scipy
I think we chose statsmodels because scipy only has 2x2 Fisher exact, so we may need to add both packages to our dependencies unless we can find another way.


ielis:
We already use scipy, so no problem with using Mann Whitney from scipy in genophenocorr.

docs/user-guide/stats.rst
```diff
@@ -0,0 +1,47 @@
...
+Fisher Exact Test (FET)
+~~~~~~~~~~~~~~~~~~~~~~~
+
+The Fisher exact test (FET) calculates the exact probability value for the
+relationship between two dichotomous variables. In our implementation, the two dichotomous variables are the genotype and the phenotype.
+For instance, the individuals of the cohort may be divded according to whether or not they have a nonsense variant and according to whether
+or not they have ataxia.
+
+
+The results of FET are expressed in terms of an exact probability (P-value), varying within 0 and 1. Two groups are
```
ielis, line 21: We use one-tailed test, right?

ielis (approved):
Perfect!

pnrobinson:
We should be using the two tailed test unless we have an hypohtesis that something is higher in a specific genotype group!

pnrobinson:
I am going to be adding more material to the FET description, Let's leave this open for a bit.

pnrobinson:
@ielis I updated the documents to explain the Fisher test. Please sanity check and then we can merge!

ielis (approved):
Fabulous!

PR #199, opened 2024-08-01
Implement `IsStructuralVariantPredicate` for testing if a variant is an SV

ielis:
Implement `IsStructuralVariantPredicate` for testing if a variant is an SV.


pnrobinson:
@ielis 
I added some code for checking specifically whether there is a chromosomal deletion. Can you please sanity check? I think we also want analogous code for duplication, inversion, and translocation. It is a bit repetitive but that is bascially it (i.e., there are very probably no more SV subtypes we want to test). If we wanted to do this like we do the HGVS variant classes, we would need an enum, but perhaps it is not worth it for just four predicates?

ielis:
The question is - do you want to test for
1. "chromosomal deletion" as defined by Sequence Ontology, or 
2. for an SV where `variant class == 'DEL'`?

Based on the need for testing for `DUP`, `INV`, etc., I suspect the option 2 is sought.

Then this would be better employed by a predicate that focuses on `variant_class` property, and can be combined with `is_structural_variant()` as:

```python
from genophenocorr.analysis.predicate.genotype import VariantPredicates

sv_del = VariantPredicates.is_structural_variant() & VariantPredicates.variant_class('DEL')
sv_inv = VariantPredicates.is_structural_variant() & VariantPredicates.variant_class('INV')
# etc..
```

where `VariantPredicates.variant_class('DEL')`  needs to be implemented.

pnrobinson:
I was trying to get both classes

"chromosomal deletion" as defined by Sequence Ontology AND  SV where variant class == 'DEL'


this would make sense if the way chromosomal deletions is represented includes both types of variant in some cohort

ielis (commented):
So, now we have a couple of new variant predicates.

We can test for `VariantClass`, `change_length`, and for `structural_type` (e.g. chromosomal deletion). This is exemplified at the end of the predicate section in the user guide.

@pnrobinson please let me know if you'd like to build convenience functions for creating predicates to test if the variant is an INV, DUP, INS, etc.. Right now, this can be done via `VariantPredicates.variant_class(VariantClass.DEL)` or similar, but as I explained, a 1bp deletion will also pass this test. Therefore, the SV tests should be built using `VariantPredicates.is_structural_variant` where we can now specify threshold (default 50bp).

We can put the convenience functions into  `VariantPredicates`.

pnrobinson (approved):
I think this should be enough to get us started. Let's merge this and use the code in some cohorts. Types of SV other than deletion are less common. If we like using the API like this then it should be easy to add a few other classes as needed.

PR #235, opened 2024-08-27
Made updates to all the notebooks to reflect the rebrand

lnrekerle:
Updated all notebooks so that they run with the rebrand from genophenocorr to gpsea. 
Also updated notebooks so that they work with latest updates, and so that we no longer have to save phenopackets and all of them are going to be pulled from phenopacket-store instead.


ielis (approved):
Looks good, thanks!

pnrobinson:
@ielis @lnrekerle 
I am approving this PR so we can move on.
However, I am noticing that we are including lots of files in the PR. I think it would be better to use the new phenopacket store toolkit to import the cohorts and to add the *.json to .gitignore ?

pnrobinson (approved):

ielis:
Hey @pnrobinson I think you misinterpreted the diff - the files are actually being *removed*, which is in line with using Phenopacket Store Toolkit.

PR #238, opened 2024-08-28
Duplicate patient

### LINKED ISSUE #227: Validate uniqueness of Patient ID
When creating a cohort, we must validate that the patient IDs are unique.

The cohort is only created if all patient IDs are unique. Otherwise we must raise an error with all issues.

lnrekerle:
closes #227 

If the patient ID already exists in the cohort, then an error will be raised that prevents the creation of a cohort, and the IDs will be printed so the user can changed the IDs.


src/gpsea/preprocessing/_patient.py
```diff
@@ -38,13 +38,20 @@ def __init__(self, patient_creator: PatientCreator[T]):
...
     def process(self, inputs: typing.Iterable[T], notepad: Notepad) -> Cohort:
         patients = []
+        patient_ids = set()
+        duplicate_pat_ids = set()
 
         for i, pp in enumerate(inputs):
             sub = notepad.add_subsection(f'patient #{i}')
             patient = self._pc.process(pp, sub)
+            if patient.patient_id in patient_ids:
+                duplicate_pat_ids.add(patient.patient_id)
```
ielis, line 48: I think it is a little better to use `labels` instead of `patient_id` because `labels` is the primary data type.

`SampleLabels` has `__eq__` and `__hash__` so we can check if the items are unique.

ielis (changes requested):
This looks great, there are just few minor things that I would like you to update.

Plus, there would be many bonus points for writing a test that shows that this actually works (although I am sure it does).

ielis (approved):
Looks great, thanks!

At this point, it is indeed hard to test the validation results apart from calling the `summarize` method and search for the expected complaints.

Let's merge..

PR #248, opened 2024-09-05
updated tqdm to print files/sec instead of it/sec

### LINKED ISSUE #241: Patients Created: 56it [00:54,  1.03it/s]
This output appears after the user has input phenopackets.
It is unclear what "56it" refers to.
It would be better to remove the "it" here

```
Patients Created: 56 [00:54,  1.03it/s]
```

lnrekerle:
Individuals Processed: 100%|██████████| 337/337 [01:38<00:00, 3.43individuals/s]

This is what the bar should look like now. 
This will close #241


ielis (changes requested):
We should probably use "individuals" as a `unit`. Phenopackets do not necessarily have to only come from a file. Take Phenopacket Store as an example. There is one ZIP archive that is open and the phenopacket data is read from the archive file on the fly..

lnrekerle:
Oh! Yah, okay, should we do individual or patient? Since the start of the bar does say "Patients Created", using patient would make sense. 

@ielis

pnrobinson:
Please alway replace the word "patient" with the work "individual".
People with rare disease want this because individual captures better the fact that affected people have lives outside of being a patient (which is a role we all play when we go to the doctor, but is not our individual identity).

lnrekerle:
@pnrobinson 
That makes a lot of sense! How does this loading bar look then?
Individuals Processed: 100%|██████████| 337/337 [01:38<00:00,  3.43individuals/s]

ielis:
@lnrekerle yes, the progress bar looks good!

ielis (changes requested):
So, the progress prompt still needs to be aligned with the prompt from the PR messages.

src/gpsea/preprocessing/_config.py
```diff
@@ -455,7 +455,7 @@ def load_phenopackets(
     # Turn phenopackets into a cohort using the cohort creator.
     # Keep track of the progress by wrapping the list of phenopackets
     # with TQDM 😎
-    cohort_iter = tqdm(phenopackets, desc="Patients Created", file=sys.stdout)
+    cohort_iter = tqdm(phenopackets, desc="Patients Created", file=sys.stdout, unit="files")
```
ielis, line 458: As far as I can see, this still says `"files"` instead of individuals, and the `desc` includes "Patients".

ielis (changes requested):
See the comment below.

src/gpsea/preprocessing/_config.py
lnrekerle: Ooof, yah I'm sorry I thought I had pushed that change when I made the comment before. It should be fixed now!

ielis (approved):
Yes, this is it.

Issue #249, opened 2024-09-06
Rename PropagatingPhenotypePredicate


pnrobinson:
PropagatingPhenotypePredicate should be renamed to HpoPredicate
(Peter has already updated the documentation page).

ielis:
Done in #251

PR #251, opened 2024-09-06
updating documentation (WIP)

pnrobinson:
I think this will need a few more iterations, but please check the current suggestion for structuring documentation. It is prob easiest to merge this soon unless we want a different structure!


ielis:
Fixes #249

pnrobinson:
@Thanks, I will merge this now since I want to focus on other things for a while. We will need to do more work on the documentation later on (any of us)!

PR #255, opened 2024-09-09
Counts table

pnrobinson:
Adds a Table that shows all variants in a cohort

@ielis 
I added two methods to VariantEffect that mix display and logic, but the alternative seemed dangerous because it would be easy to forget to include a new code etc. Please check this and change if there is a better way
see
```
@staticmethod
def to_display(veffect: "VariantEffect") -> str:
```
and
```
@staticmethod
def structural_so_id_to_display(so_term: typing.Union[TermId, str]) -> str:
```


ielis (dismissed):
Looks great, thank you!

I took the liberty to make few adjustments. Most importantly, I changed the name to `gpsea.view.CohortVariantViewer`.

I also added a Tutorial section for showing the generated table.

Please merge when convenient!

ielis (approved):
Now, that the tests actually pass, this looks good.

PR #260, opened 2024-09-11
adding imprecise SVs

pnrobinson:
Addresses https://github.com/monarch-initiative/gpsea/issues/259

Not entirely clear to me what the testing is for functionality. For instance, why do we want this function to return None? Please document! :-0

```
    def test_find_large_structural(
        self,
        fpath_test_genomic_interpretations: str,
        pp_vc_finder: PhenopacketVariantCoordinateFinder,
    ):
        fpath_pp = os.path.join(fpath_test_genomic_interpretations, 'chromosomal_deletion.ANKRD11.json')
        gi = read_genomic_interpretation_json(fpath_pp)

        vc = pp_vc_finder.find_coordinates(gi)
        assert vc is None
```


src/gpsea/model/_variant.py
```diff
@@ -242,6 +242,11 @@ class VariantClass(enum.Enum):
     A breakend.
     """
 
+    TRANSLOCATION = 7
```
ielis, line 245: Isn't this redundant with `BND` above?

src/gpsea/preprocessing/_phenopacket.py
```diff
@@ -528,6 +534,10 @@ def _map_structural_type_to_variant_class(
                 return VariantClass.DEL
             elif structural_type.id in self._so_duplications:
                 return VariantClass.DUP
+            elif structural_type.id in self._so_translocations:
+                return VariantClass.TRANSLOCATION
```
ielis, line 538: Depending on my comment in `_variant.py`, this may need to be changed to `VariantClass.BND`.

ielis (changes requested):
Hi @pnrobinson,

thank you for pointing this out. The topic of breakends/translocations has not yet been addressed by GPSEA and it seems like there *is* a need.

The PR looks OK, there are few things that I'd like to clarify though. Please see below.

src/gpsea/preprocessing/_phenopacket.py
pnrobinson: I think BND is broader than just translocation ("BND can be used to represent reciprocal translocations or non-reciprocal break ends. ") but in any case, BND is a technical description rather than a biological one, and so I do not think we should actually keep BND as a Variant Category

ielis: So do you recommend renaming `BND` to `TRANSLOCATION`?

pnrobinson: That would work, yes, I am guessing it has not been used in any other context yet (if at al!)!

ielis: It has not been used, and I kind of planned to use `BND` as a translocation. Let's then keep `TRANSLOCATION` and remove `BND`.

pnrobinson:
@ielis Using TRANSLOCATION instead of BND seems good. Probably we can merge this?

ielis (approved):
Awesome!

PR #266, opened 2024-09-12
Fix Uniprot API bug

ielis:
Our Uniprot service was not able to get metadata for some protein IDs, such as `NP_001365381.1` of *ITPR1* gene. 

In some cases, the REST response include only one result. If this happens, we will extract the metadata from the response even if there is not an exact match between the query protein ID and the cross-references included in the response.

### DIFF
--- a/src/gpsea/analysis/predicate/genotype/_predicates.py
+++ b/src/gpsea/analysis/predicate/genotype/_predicates.py
@@ -359,20 +359,20 @@ class StructuralTypePredicate(VariantPredicate):
 
 
 def _decode_operator(op: str) -> typing.Callable[[int, int], bool]:
-        if op == '<':
-            return operator.lt
-        elif op == '<=':
-            return operator.le
-        elif op == '==':
-            return operator.eq
-        elif op == '!=':
-            return operator.ne
-        elif op == '>=':
-            return operator.ge
-        elif op == '>':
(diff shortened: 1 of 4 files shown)


pnrobinson:
@ielis 
This is probably not what we want to do because the numbering will be incorrect!

ielis:
What numbering?

pnrobinson:
This is the typical problem

reported as NM_001168272.2:c.4459_4460delinsGA; 
correct HGVS is 
NM_001168272.2:c.4459_4460inv=NP_001161744.1:p.(S1487D);
NM_001378452.1:c.4504_4505inv=NP_001365381.1:p.(S1502D)


So for the same variant, we have different coordinate systems depending on the transcript.
It would be better to define simple schema and allow the user to add the information themselves but with the correct transcript. So if we failed, we print a BIG RED ERROR MESSAGE and explain what to do!

ielis:
Then the question is the following:

In case we query UNIPROT with `NM_001168272.2` (the 1st tx from your example) and we get back a response with 1 result only, are we willing to use it if the response does not *explicitly* include the cross-reference to `NM_001168272.2`?

I think this is not completely unreasonable, since Uniprot returned a response with data for just *one* protein sequence.

@pnrobinson what are your thoughts?

PR #273, opened 2024-09-18
DataFrame for protein features

pnrobinson:
@ielis 
Could you give me quick feedback if this is OK?
The user will pass a dataframe and get a ProteinMetaData
I would additionally add the region type as an additional column.

This is useful for cases in which UniProt does not have the actual domains of interest (e.g. ITPR1) as well as for cases where the API does not work.

### DIFF
--- a/src/gpsea/model/_protein.py
+++ b/src/gpsea/model/_protein.py
@@ -3,9 +3,9 @@ import enum
 import typing
 
 import hpotk
+import pandas as pd
 
 from .genome import Region
-
 
 class FeatureInfo:
     """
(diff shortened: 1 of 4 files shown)


ielis:
I think this is OK in principle.

However, when thinking about how we use it, we may want to just create `ProteinMetadata` from the data frame directly, without implementing `ProteinMetadataService`. So, something like this:

```python
class ProteinMetadata:  # The existing class ...

    # ... gets a new static constructor
    @staticmethod
    def from_feature_frame(
        protein_id: str,
        label: str,
        features: pd.DataFrame,
        length: int,
) -> "ProteinMetadata":
        # TODO: implement
        pass
```

Then, as a next step to take advantage of the code above, we will have to update the protein-dependent predicates [here](https://github.com/monarch-initiative/gpsea/blob/a6caa22891e4e957006141c95e26b225b7424d0f/src/gpsea/analysis/predicate/genotype/_predicates.py#L523) and [here](https://github.com/monarch-initiative/gpsea/blob/a6caa22891e4e957006141c95e26b225b7424d0f/src/gpsea/analysis/predicate/genotype/_predicates.py#L583) to take the `ProteinMetadata` of interest directly, instead of the service.

ielis (changes requested):
OK in principle, but see my comment above regarding the TODO.

pnrobinson:
sounds good. Should I also just add an annotate function (which is required) that throws a NotImplemented error?

pnrobinson:
@ielis I implemented your suggestion and added a test class, seems to work.

ielis:
Yeah, it seems to be OK. Let me have a closer look, I may reformat or rearrange some things. Then, as long as you're done with this, I'll merge.
Pls give me a :+1: if it sounds OK.

ielis (approved):
Awesome!

PR #277, opened 2024-09-21
Uniprot JSON prototype

pnrobinson:
Some opaque error with TestUniprotJsonToMetadata, but the function is working.
Note that we need to extend the list of protein features, e.g., coiled coil occurs in the test example but is not in our list of features.


pnrobinson:
Fixed unittest (a "self" was missing in the function signature).
This PR adds a function to create a ProteinMetadata from the JSON file that can be downloaded from the UniProt website, and is a backup in case the API does not work. I also added a test class and a test-downloaded UniProt JSON file. I needed to add two FeatureTypes to our list because they were missing -- we should probably check whether there are more feature types we need to add.

pnrobinson:
btw it is easy to access the entire JSON file using an option on the website, e.g., 
https://rest.uniprot.org/uniprotkb/P37173
I suspect that we are simply not parsing the JSON file correctly? because the correct NP is is present for the gene I am working on.

ielis (approved):
Awesome!

PR #279, opened 2024-09-23
protein domain viz documentation

pnrobinson:
This PR adds documentation for how to add protein domains manually with an explanation of why this might be useful.


pnrobinson:
@ielis We still need documentation about how to use these domains for analysis, but that should be a separate PR. Can we merge this?

ielis:
Yes, the documentation regarding *how to use all this* indeed needs to be written. The sections that exemplify the usage include, but are not limited to:

- cohort exploration - plot protein diagram, show protein feature table
- predicates - example of a GT predicate for testing if a variant overlaps with a certain protein feature

Regarding merging, we could merge it, but I would appreciate if you could go through the docs and improve the readability if you find anything odd.

ielis (approved):
I think the structure is now OK. The text readability could, however, be further improved.

Issue #284, opened 2024-09-25
Simplify AR predicate


ielis:
The AR predicate is exhaustive in terms of produced categorizations: `HOM_REF`, `HET`, and `BIALLELIC_ALT`.

In practice, however, it is not so useful to differentiate between `HOM_REF` and `HET`. Therefore, the predicate should be simplified to categorize the individuals into 2 categories only:

- `HOM_REF`, `HET`
- `BIALLELIC_ALT`

pnrobinson:
Please let's not use HOM_REF, because in our intended use case we are not comparing to REF, we are comparing to other variants. I think it would be better to require users to name the categories to avoid confusion!

PR #289, opened 2024-09-26
Measurements

### LINKED ISSUE #253: t test: CYP21A2
Use the CYP21A2 cohort (on the develop branch of phenopacket-store) to demonstrate the t test functionality.

ielis:
Add measurements and implement T test for quantitative tests.

Closes #253


ielis:
One thing that is not entirely clear to me is what can we do with this variant:
[Individual 59](https://github.com/monarch-initiative/phenopacket-store/blob/5738cf9eeb00ea23e6fa765183b24fbaa04a9822/notebooks/CYP21A2/phenopackets/PMID_30968594_individual_59.json#L198)

image

The SO term only informs us about "something" in the promotor, but it is not clear what it is (DEL, DUP, ...).

pnrobinson:
@ielis 
You are right, we need to extend the promoter data type to allow annotation to the genomic variant. I will do this.

pnrobinson:
This issue cannot be closed until https://github.com/monarch-initiative/gpsea/issues/291 is finished.

ielis:
In addition to #291 , this also depends on and will be fixed by #284 .

ielis:
@pnrobinson the code is ready to go and I will merge it.
I will not wait for resolution of #291 because at this point we just ignore the weird promoter variant and the code will "just work" as soon as we fix the offending phenopacket.

pnrobinson:
@ielis The problem was that the promoter variants only had invalid HGVS, c.-500A>G, but -500 lies outside of the transcript and so according to HGVS we need to provide genomic coordinates, but the authors provide next to no context so it would be difficult to know exactly what they meant. Anyway, I had forgotten to add the "description", which I now do in pyphetools, when we rerun the cohort it will have another field to make the variants unique. I am going to add this to pyphetools and add a test or two to fix the weirdness above.

pnrobinson:
this was just a minor bug in the script,  the library code is correct.

PR #292, opened 2024-09-28
adding improvements to protein variant table and adding documentation.

pnrobinson:
Adding list of variants to output table. Also adding verbiage to exploratory about why and how to use the table.


ielis:
@pnrobinson this looks awesome!

I will turn this into a testable guide section and then we can close it.

ielis:
The CI is failing because the measurements guide cannot handle 3 GT categories at this time.
This depends on and will be fixed by #284

ielis (approved):
Awesome!

PR #297, opened 2024-10-02
Found bug if there are no protein features. 

lnrekerle:
If a protein has no features, an error returned when creating a ProteinVariantViewer. 
Fixed it by skipping if hgvs_p is None. It now prints an empty table.


pnrobinson (approved):
+1

PR #298, opened 2024-10-04
Remove notebooks

pnrobinson:
@ielis any objection to moving all of the notebooks to gpsea-cs?


ielis:
No objection at all, as long as the PR moves _all_ notebooks to gpsea-cs.

PR #310, opened 2024-10-13
Sort variant effects table and report percentages

pnrobinson:
Adds a column with percentages to the variant effects table and sorts the column according to counts.


ielis (approved):
Looks good!

PR #328, opened 2024-10-22
Counting male and female individuals for CohortViewer

pnrobinson:
@ielis 
I wonder if we can also add information about the number of patients with age of onset etc here. I cannot find accessors in the Patient class though -- is this possible?


ielis:
* `Patient` has `age` and `vital_status` optional properties.
* `Disease` has an optional `onset` property.
* `Phenotype` has also an optional `onset` property.

These should be consulted when creating a summary. Maybe we need few convenience methods on `Patient`?

pnrobinson:
Somehow I do not see where this information is in the Patient class?

ielis:
I am attaching links:

* Patient has [age](https://github.com/monarch-initiative/gpsea/blob/3c0ff7c68016cb84b5ef337ae20c831a851278e6/src/gpsea/model/_cohort.py#L146) and [vital_status](https://github.com/monarch-initiative/gpsea/blob/3c0ff7c68016cb84b5ef337ae20c831a851278e6/src/gpsea/model/_cohort.py#L153) optional properties.
* Disease has an optional [onset](https://github.com/monarch-initiative/gpsea/blob/3c0ff7c68016cb84b5ef337ae20c831a851278e6/src/gpsea/model/_phenotype.py#L190) property.
* Phenotype has also an optional [onset](https://github.com/monarch-initiative/gpsea/blob/3c0ff7c68016cb84b5ef337ae20c831a851278e6/src/gpsea/model/_phenotype.py#L100) property.

pnrobinson:
@ielis 
locally I get this error:
```python
ImportError while loading conftest '/Users/robin/GIT/gpsea/tests/conftest.py'.
conftest.py:8: in <module>
    from gpsea.analysis.predicate.genotype import (
E   ImportError: cannot import name 'allele_count' from 'gpsea.analysis.predicate.genotype' (/Users/robin/GIT/gpsea/venv/lib/python3.12/site-packages/gpsea/analysis/predicate/genotype/__init__.py)

Process finished with exit code 4

Empty suite
```
I do not think it is related to my PR but please check.

This PR adds a function list_measurements to the model - cohort class, and updates the cohort (Python and HTML) in view in order to show a few additional items in the CohortViewer summary.

ielis (approved):
I updated the HTML reports that are embedded into the documentation and they looked OK.

The code looks OK too. However, with the growing amount of the information that we report via HTML items, we will need to do a dedicated pass through the `gpsea.view` code & reports (not in scope of this PR).

Please merge at your discretion.

pnrobinson:
Thanks, let's merge now for simplicity. We will need to go over several notebooks together and update the documentation. For now, AFAIK, there are three cohorts that still need curation, this will take another week or so, and then we should fionalize the phenopacket-store version and start with the final analysis.

Issue #344, opened 2024-11-11
"Zinc finger"


pnrobinson:
The protein display is not showing domains of type "Zinc finger", 
example ZFX 

https://www.uniprot.org/uniprotkb/P17010/entry#family_and_domains

@ielis 
We can consider throwing an error if there is a domain that we do not recognize, but I think in the early stages of this class we incorrectly assumed that there would be just a small number of domain types. At the very least we need to show the user there was a problem. I will add this for now.

ielis:
Sounds good.

ielis:
Fixed by #346

PR #346, opened 2024-11-12
adding zinc finger to ProteinFeatures

pnrobinson:
@ielis 
There are many errors related to allele_count. These are not related to this PR. Is this a known issue or should I try to fix?


ielis:
I do not see any issue with `allele_count` in the context of this PR. The CI fails because of a REST API timeout, which is a common issue. 
If you see 

```
UNEXPECTED EXCEPTION: HTTPError('429 Client Error: TOO MANY REQUESTS for url: https://rest.variantvalidator.org/VariantValidator/tools/gene2transcripts/NM_181486.4')
```
then pls just re-run the CI until it passes.

pnrobinson:
Can you please check this locally? Why I run pytest in Pycharm I get about 20 errors because of this.

ielis:
One CI ran OK (3.11), which should mean that the issue is local to your environment. The CI run for 3.10 stopped due to timeout, leading to cancellation of 3.12.

So, thanks to 3.11 finishing, I am almost sure that the PR will pass on my side.

ielis:
Fixes #344

PR #352, opened 2024-11-13
Devries score update

pnrobinson:
@ielis 

I updated the heuristic according to our discussion with Bert and Lex, and also updated the documentation. It should be ready to go.

### DIFF
--- a/src/gpsea/analysis/pscore/_hpo.py
+++ b/src/gpsea/analysis/pscore/_hpo.py
@@ -166,7 +166,7 @@ class DeVriesPhenotypeScorer(PhenotypeScorer):
     ) -> float:
         """
         Calculate the dev delay component of the score
-        
+
         Args:
             observed_term_ids: terms observed in patient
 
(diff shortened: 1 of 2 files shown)


src/gpsea/analysis/pscore/_hpo.py
```diff
@@ -216,22 +216,21 @@ def _term_or_descendant_count(
...
-            the total count of the terms equal to or descending from the target_tid
+            1 if at least one term is equal to or descending from the target_tid, otherwise 0
         """
-        total_count = 0
         for term_id in observed_term_ids:
             for desc_tid in self._hpo.graph.get_ancestors(term_id, include_source=True):
                 if desc_tid.value == target_tid:
-                    total_count += 1
-        return total_count
+                    return 1
```
ielis, line 224: @pnrobinson 
note that `_term_or_descendant_count` and `_term_or_descendant` methods are now functionally identical. Is that on purpose?

ielis (changes requested):
@pnrobinson looks good but let's first clarify my comment below.

PR #357, opened 2024-11-15
also return t statistic (not just the p value)

pnrobinson:
@ielis 
See https://github.com/monarch-initiative/gpsea/issues/356


pnrobinson:
docs/user-guide/input-data.rst: 45 warnings
  /opt/hostedtoolcache/Python/3.11.10/x64/lib/python3.11/site-packages/matplotlib/transforms.py:2048: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
    self._mtx[0, 2] += tx


Does this mean that returning X, Y is no longer allowed? Do we need parens?

PR #359, opened 2024-11-18
Adding TRANSMEMBRANE and TOPOLOGICAL DOMAIN to protein types 

pnrobinson:
We need this to display in the protein graphic. 
@ielis I am not sure we want to crash if there is an unrecognized domain. Maybe it is better to just call anything we do not recognized "Region" and to emit a warning with a request for feedback? I am pretty sure we did not get all possible domains and I have not been able to find a complete list.


ielis (approved):
The PR looks OK, of course we must add the missing feature types.

However, we should also re-consider if defining `FeatureType` as an enum is actually justified. Perhaps we should change it to a `str`? This would affect one predicate (`VariantPredicates.protein_feature_type()`) and otherwise probably not too much other stuff.

Let's merge this PR and re-evaluate if a need for a new protein `FeatureType` arises.

PR #384, opened 2024-12-21
Removing erroneous X-axis label (# Codons)

pnrobinson:
The X-axis label was erroneous. Our users will know that the X axis represents amino acid residues, and nearly all of the plots you see like this do not have any label at all, so I am just removing it.


ielis (approved):
Looks great, thanks!

PR #408, opened 2025-01-21
Add numtest and use doctest.ELLIPSIS_MARKER to fix doctest errors when doing floating point number comparisons

justaddcoffee:
doctests in survival.rst and phenotype_scores.rst are failing for me on my Macbook M1 MacOS v12.5

```
FAILED               [  2%]
docs/user-guide/analyses/survival.rst:0 ([doctest] survival.rst)
120 
121 We execute the analysis by running
122 
123 >>> result = survival_analysis.compare_genotype_vs_survival(
124 ...     cohort=cohort,
125 ...     gt_clf=gt_clf,
126 ...     endpoint=endpoint,
127 ... )
128 
129 >>> result.pval
Expected:
    0.06200425830044376
Got:
    0.06200425830044377

/Users/jtr4v/PythonProject/gpsea/docs/user-guide/analyses/survival.rst:129: DocTestFailure
```

This PR adds numtest as a test dependency and then uses ELLIPSIS_MARKER to allow approximate comparisons

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -3,6 +3,9 @@ import doctest
 import os
 import sys
 
+# use numtest to allow floating point comparisons
+doctest.ELLIPSIS_MARKER = '...'
+
 # -- General configuration ------------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -3,8 +3,6 @@ import doctest
 import os
 import sys
 
-# use numtest to allow floating point comparisons
-doctest.ELLIPSIS_MARKER = '...'
 
 # -- General configuration ------------------------------------------------
 
--- a/docs/user-guide/analyses/phenotype-scores.rst
+++ b/docs/user-guide/analyses/phenotype-scores.rst
@@ -39,7 +39,7 @@ This is a non-parametric test that compares the medians of the two classes to de
 >>> r = stats.mannwhitneyu(x=class1, y=class2, alternative = 'two-sided')
 >>> p_value = r.pvalue
 >>> float(p_value)
-6.348081479150e...-06
+6.348081479150...e-06
 
 
 p value of `6.348081479150e-06` suggests a significant difference between the classes.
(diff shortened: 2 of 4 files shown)


docs/user-guide/analyses/phenotype-scores.rst
```diff
@@ -38,11 +38,11 @@ This is a non-parametric test that compares the medians of the two classes to de
 >>> class2 = [4, 5, 3, 4, 3, 3, 3, 4, 4, 5, 5, 2, 3, 0, 3, 5, 2, 3]
 >>> r = stats.mannwhitneyu(x=class1, y=class2, alternative = 'two-sided')
 >>> p_value = r.pvalue
->>> float(p_value)
-6.348081479150902e-06
+>>> float(p_value) # doctest: +ELLIPSIS
```
ielis, line 41: The doctest runner seems to recognize the `...` even without `#doctest: +ELLIPSIS` marker, and the test passes here. Does the same happen on your end? If yes, pls remove `# doctest: +ELLIPSIS`.

docs/user-guide/analyses/phenotype-scores.rst
```diff
@@ -265,8 +265,8 @@ In case of the *RERE* cohort, the analysis shows a significant difference
 between the number of structural defects in individuals
 with point vs. loss-of-function mutations.
 
->>> result.pval
-0.012074957610483744
+>>> result.pval # doctest: +ELLIPSIS
```
ielis, line 268: Same here, the test passes even without the ELLIPSIS marker.

docs/conf.py
```diff
@@ -3,6 +3,9 @@
 import os
 import sys
```
ielis, line 5: AFAIK, this is the default ellipsis marker. The declaration is, therefore, redundant and should be removed.

pyproject.toml
```diff
@@ -52,12 +52,13 @@ dynamic = ["version"]
 test = [
     "pytest>=7.0.0,<8.0.0",
     "pytest-cov",
+    "numtest"
```
ielis, line 55: I am not sure we really need this dependency. All tests passed on my end *without* `numtest`, just with the `...` at proper places. Unless I am missing something, we do not need `numtest`, and by applying Occam's razor, it should be removed.

ielis (changes requested):
Hello @justaddcoffee,

thanks a lot, using ellipsis in float comparisons is a very good and useful idea. I have a few comments & questions, can you pls see below?

Thank you!

justaddcoffee:
Thanks @ielis - you are right, the `numtest` dependency,` doctest.ELLIPSIS_MARKER = '...'` and `DOCTEST +ELLIPSIS annotations` are not necessary here, so I've removed them. It looks like all this ELLIPSIS stuff is now default, and I was reading some stale documentation

ielis (approved):
Looks great now, thank you!

